PR #61, opened 2025-02-03
check process before getting its state

GuzekAlan:


GuzekAlan (commented):
Currently these errors are shown: 
image
 But these are our own errors. I don't know if we want to show them or not.

GuzekAlan:
> Currently these errors are shown: image But these are our own errors. I don't know if we want to show them or not.

After discussion we don't want to show any errors if this behaviour is expected

kraleppa (approved):

PR #68, opened 2025-02-04
Made events more secure

kraleppa:
- added rate limiting
- added events limit (when new appears above this limit, old are deleted)
- added start/stop button


GuzekAlan:
Looks good but when a lot of events are pushed stopping occurs after some time

PR #70, opened 2025-02-05
Fixed events tracing after debugger crash

kraleppa:


lib/live_debugger/services/callback_tracing_service.ex
```diff
@@ -168,11 +168,17 @@ defmodule LiveDebugger.Services.CallbackTracingService do
...
-    n
-    |> Trace.new(module, function, args, pid)
-    |> do_handle(recipient_pid, ets_table_id, n)
+    if function in CallbackUtils.callbacks_functions() do
+      n
+      |> Trace.new(module, function, args, pid)
+      |> do_handle(recipient_pid, ets_table_id, n)
+    else
+      n
+    end
```
GuzekAlan, line 177: Maybe add some info/warning here that some unknown function callback was traced?

GuzekAlan (approved):

PR #74, opened 2025-02-10
Add nested live view support

GuzekAlan:


GuzekAlan:
We can wait with this PR and return to it after the refactor of callback traces

PR #83, opened 2025-02-13
collapse tree on specific level when number of children is too big

GuzekAlan:


lib/live_debugger/components/tree.ex
```diff
@@ -20,6 +22,13 @@ defmodule LiveDebugger.Components.Tree do
   attr(:class, :string, default: nil, doc: "CSS class")
 
   def tree(assigns) do
+    assigns =
+      assign(
+        assigns,
+        :max_opened_node_level,
+        max_nesting_level(assigns.tree_node, @max_node_number)
```
kraleppa, line 29: This type of calculation shouldn't be probably triggered during render - if it is hard to move we can leave it as it is and refactor it later (please leave a comment here and create a task)

kraleppa (approved):

PR #89, opened 2025-02-17
Change assigns section design

GuzekAlan:


kraleppa (commented):
Small thing - on fullscreen mode there is no gap between callback argument
Screenshot 2025-02-17 at 17 48 36

GuzekAlan:
I've checked now and there is gap 
image

Maybe assets didn't recompile and class for gap wasn't there when you've tested it @kraleppa

kraleppa (approved):

PR #90, opened 2025-02-17
Redesigned LiveSessions view

kraleppa:


lib/live_debugger/live_views/sessions_dashboard.ex
```diff
@@ -18,50 +18,55 @@ defmodule LiveDebugger.LiveViews.SessionsDashboard do
...
-            <.icon phx-click="refresh" name="icon-refresh" class="text-primary cursor-pointer" />
+          <div class="flex gap-4 items-center justify-between">
+            <div class="text-primary font-semibold text-2xl">Active LiveSessions</div>
+            <.button phx-click="refresh" variant="outline">
+              <div class="flex items-center gap-2">
+                <.icon name="icon-refresh" class="w-4 h-4" />
+                <p>Refresh</p>
+              </div>
+            </.button>
           </div>
```
GuzekAlan, line 41: Could you move this outside (so that spinner will only cover table/list when reloading)

GuzekAlan (commented):
There is no gap when list is shown:
image

lib/live_debugger/components.ex
```diff
@@ -306,6 +306,57 @@ defmodule LiveDebugger.Components do
...
+    ~H"""
+    <div class={["p-4 bg-white rounded" | List.wrap(@class)]}>
+      <table class="w-full">
+        <thead class="border-b border-blue-100">
+          <tr class="text-left text-primary text-sm h-11 mx-16">
+            <th :for={col <- @column} class="first:pl-2"><%= col.label %></th>
+          </tr>
+        </thead>
+        <tbody class="text-primary text-sm text-bold">
+          <tr :for={row <- @rows} class="h-11 hover:bg-blue-50">
```
GuzekAlan, line 326: could you please update colours in `tailwind.config.js`? As I see the names changed and previous one are a little bit outdated to figma ( `primary-500` is `swm-blue-950`). I think that this blue is not the sam as on designs.

assets/tailwind.config.js
```diff
@@ -22,6 +22,8 @@ module.exports = {
           DEFAULT: '#001A72',
           5: '#F7F9FF',
           20: '#E6ECFE',
+          50: '#F1F4FF',
+          100: '#E6ECFE',
```
GuzekAlan, line 26: 100 is same as 20. can you please update `primary-20` to `primary-100`?

lib/live_debugger/components.ex
```diff
@@ -306,25 +306,38 @@ defmodule LiveDebugger.Components do
     """
   end
 
-  attr(:rows, :list, default: [])
+  attr(:rows, :list, default: [], doc: "Elements that will be displayed in the list")
   attr(:class, :any, default: nil, doc: "Additional classes.")
+  attr(:on_row_click, :string, default: nil)
+  attr(:row_click_target, :any, default: nil)
+  attr(:row_click_key, :atom, default: nil)
```
GuzekAlan, line 313: How about passing function instead of key (it will be more general and can apply modifications of values or passing multiple values). Or add functionality to pass list of keys

GuzekAlan (approved):

PR #92, opened 2025-02-18
Callback trace redesign

GuzekAlan:


lib/live_debugger/components.ex
```diff
@@ -90,6 +90,7 @@ defmodule LiveDebugger.Components do
   @doc """
   Collapsible element that can be toggled open and closed.
   It uses the `details` and `summary` HTML elements.
+  If you add `hide-on-open` class to element it will be hidden when collapsible is opened.
```
kraleppa, line 93: Maybe some example?

kraleppa (approved):

PR #99, opened 2025-02-20
Split dev assets and prod assets

hhubert6:


.gitignore
```diff
@@ -31,3 +31,4 @@ live_debugger-*.tar
 
 node_modules/
 
+/priv/static/dev/
```
kraleppa, line 34:
```suggestion
/priv/static/dev/

```

config/config.exs
```diff
@@ -11,6 +11,12 @@ if config_env() == :dev do
         ~w(js/app.js --bundle --minify --sourcemap=external --target=es2020 --outdir=../priv/static/),
       cd: Path.expand("../assets", __DIR__),
       env: %{"NODE_PATH" => Path.expand("../deps", __DIR__)}
+    ],
+    dev: [
```
kraleppa, line 15: I'd change naming to e.g. `deploy_build` and `dev_build` to emphasize that it is executed by bot after merge

config/config.exs
```diff
@@ -23,6 +29,15 @@ if config_env() == :dev do
       --minify
     ),
       cd: Path.expand("../assets", __DIR__)
+    ],
+    live_debugger_dev: [
```
kraleppa, line 33: same here: `dev_build` and `deploy_build`

lib/live_debugger/endpoint.ex
```diff
@@ -15,6 +15,10 @@ defmodule LiveDebugger.Endpoint do
     longpoll: true
   )
 
+  if Mix.env() == :dev do
+    plug(Plug.Static, at: "/assets", from: {:live_debugger, "priv/static/dev"}, gzip: false)
+  end
```
kraleppa, line 20: I'd add `else` here

kraleppa (commented):
Also, sometimes developer wants to use `mix assets.build` on the dev mode (the same command which Github bot uses for building assets). 

I'd be good to parematrize this alias so we have e.g. `mix assets.build dev` or `mix assets.build deploy`

This will have to be changed in README too (see contributing section)

lib/live_debugger/endpoint.ex
kraleppa: Oh, I forgot about `client.js` - it should be served both in `dev` and `prod`.

kraleppa (approved):
Approved ✅ But please wait with merging till we release `v0.1.0

PR #104, opened 2025-02-24
Add visual separator between "historical" and "just-loaded events"

hhubert6:


assets/js/app.js
```diff
@@ -53,6 +53,28 @@ let liveSocket = new LiveSocket('/live', Socket, {
...
+  let separator = document.querySelector('#separator');
+
+  if (separator) {
+    tracesListDiv.removeChild(separator);
+  } else {
+    separator = document.createElement('div');
+    separator.id = 'separator';
+    separator.innerHTML = `
+      Historical events
+      <div class="border-b h-0 border-primary-100"></div>
```
kraleppa, line 69: This needs some UI polishing - I was thinking about sth like `unread messages` separator in telegram, but you can try asking Maciej the designer if he could come up with something 
Screenshot 2025-02-24 at 14 00 52

kraleppa (commented):
Also, when I clear traces I still have separator - imo it'd be good if we displayed standard info when there are no traces

https://github.com/user-attachments/assets/4eb3d247-8211-474a-8ac6-6c0f89b708b0

assets/js/app.js
```diff
@@ -53,6 +53,38 @@ let liveSocket = new LiveSocket('/live', Socket, {
...
+  let separator = document.querySelector('#separator');
+
+  if (separator) {
+    tracesListDiv.removeChild(separator);
+  } else {
+    separator = document.createElement('div');
+    separator.id = 'separator';
+    separator.innerHTML = `
+      <div class="border-b h-0 border-primary-100"></div>
+      <div class="px-6 py-1 text-primary font-normal text-center text-xs">Historical events</div>
```
kraleppa, line 69: why don't you add `border-y` to the middle `div`? It should probably do the trick

assets/js/hooks/trace_list.js
```diff
@@ -0,0 +1,56 @@
...
+      separator.id = 'separator';
+      separator.innerHTML = `
+          <div class="h-6 my-1 font-normal text-xs text-secondary-600 flex align items-center">
+            <div class="border-b border-secondary-200 grow"></div>
+            <span class="mx-2">Past Traces</span>
+            <div class="border-b border-secondary-200 grow"></div>
+          </div>
+        `;
+      return separator;
+    };
```
GuzekAlan, line 14: Creating component inside JS hook seems a little bit weird. Have you considered just appending info to stream after fetching past callbacks (in `:fetch_existing_traces`)? If there is a reason to do it that way let me know

hhubert6: It's very tricky to use just stream to do that. I couldn't make it work but I can try move html markup away from the hook.

GuzekAlan (approved):
Nice 😌

lib/live_debugger/live_components/traces_list.ex
```diff
@@ -223,6 +247,20 @@ defmodule LiveDebugger.LiveComponents.TracesList do
     """
   end
 
+  attr(:id, :string, required: true)
+
+  def separator(assigns) do
```
kraleppa, line 252: It can be private I guess

PR #148, opened 2025-02-27
Refactor nested-badge-component

hhubert6:


lib/live_debugger/components.ex
```diff
@@ -462,23 +462,14 @@ defmodule LiveDebugger.Components do
     """
   end
 
-  def nested_badge(assigns) do
+  attr(:text, :string, default: "Nested")
+  attr(:icon, :string, default: "icon-nested")
```
kraleppa, line 466: Tbh I'd delete this default values and make these required - this should be generic component

PR #149, opened 2025-02-27
Layout update

bkilshaw:
Potential tweaks for the layout

1. Move module and PID to a cardless header at the top, saves space
2. Stack assigns and callback traces vertically instead of having them side by side.
   - This tends to be a better use of the space (ie. If the callbacks block is longer than assigns, everything under assigns is wasted space)
   - This gives us more horizontal space for longer lines
 3. Removed max width on container so we can use all the space available
 4. Adds labels and dividers to the args in the callback blocks so it's more obvious what's what

CleanShot 2025-02-26 at 11 11 43@2x


kraleppa:
Hey, we'll be making major updates to the UI very soon that overlaps with your changes. 

Thanks for the contribution we'll consider your suggestions when polishing the UI, but for now we won't be able to merge your changes - I'll close the PR

Issue #152, opened 2025-03-01
[BUG] LiveDebugger JS file not loading properly


woodward:
### Description

The LiveDebugger JS files are not loading for some security reason.

### Environment

* LiveDebugger 0.1.0
* elixir 1.18.2-otp-27
* Phoenix 1.7.20
* Phoenix LiveView 1.0.4
* Operating system: MacOS
* Browser(s):  Chrome

### Actual behavior

If you open the browser dev tools in Chrome and navigate to the Console tab, you see this error:

```
Refused to load the script 'http://127.0.0.1:4007/assets/client.js' because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-inline' https://cdn.tailwindcss.com". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
```
Note that we have the following line in our `root.html.heex` layout:

```
    <%= if Application.get_env(:live_debugger, :browser_features?) do %>
      <script id="live-debugger-scripts" src={Application.get_env(:live_debugger, :assets_url)}>
      </script>
    <% end %>
```

### Expected behaviour

No JS errors.

Thanks for your help with this - we're really loving LiveDebugger!  It's already helped us find a couple of bugs in our LiveView app. 

-- Greg

kraleppa:
Hey @woodward, great to hear that LiveDebugger is doing its job!

When it comes to the error - what I see here is that your CSP configuration is blocking fetching LiveDebugger JS. It's because LiveDebugger runs on a different port than your actual app.

You probably have a similar plug in your app: 

```elixir
    plug(
      :put_secure_browser_headers,
      %{
        "content-security-policy" =>
          "script-src 'self' 'unsafe-inline' https://cdn.tailwindcss.com"
      }
    )
```

This plug sets the HTTP Content-Security-Policy header to specify that scripts can only be sourced from the same origin (`self`), can include inline scripts (`unsafe-inline`), and can be loaded from `https://cdn.tailwindcss.com`. 

To fix the issue you need to extend your policies to allow fetching JS files from the port of LiveDebugger, like this:

```elixir
      %{
        "content-security-policy" =>
          "script-src 'self' 'unsafe-inline' https://cdn.tailwindcss.com http://127.0.0.1:4007"  
      }
```

**Warning!**

This config should only be applied in `dev` environment. If your app is used on production you can conditionally change values for CSP headers like here

```elixir
  @content_security_policy (case Mix.env do
[...]

PR #159, opened 2025-03-03
Add a comment to the README about how to set up access to the LiveDeb…

woodward:
See the corresponding GitHub issue:
https://github.com/software-mansion/live-debugger/issues/152


woodward:
Feel free to resolve this however you see fit; I just thought it would be good to mention this item somehow in the docs, but it sounds like you've got a solution that you would prefer.  I just thought it would be good to be in the docs in some form or fashion.

PR #163, opened 2025-03-04
Task: fixed streams in callback traces part. I

kraleppa:
This PR is associated with moving modal rendering out of stream - it improves rendering by a lot, but it is still not perfect (there will be a continuation in a following PR)

Also, I refactored hooks - you can ignore all of them except Fullscreen hook which was changed. The rest of them is copy paste


assets/js/hooks/fullscreen.js
```diff
@@ -0,0 +1,29 @@
...
+    this.el.addEventListener(`${this.el.id}-close`, this.handleClose);
+
+    // Events from the server
+    this.handleEvent(`${this.el.id}-open`, this.handleOpen);
+    this.handleEvent(`${this.el.id}-close`, this.handleClose);
+  },
+  destroyed() {
+    this.el.removeEventListener(`${this.el.id}-open`, this.handleOpen);
+    this.el.removeEventListener(`${this.el.id}-close`, this.handleClose);
+  },
```
GuzekAlan, line 26: This does not work with closing modal with `ESC` button:
image

`handleClose` and `handleClosed` were separated earlier

kraleppa: It seems that it breaks only in case of assigns when closing with `esc` - it's working fine for me with callback traces, correct?

lib/live_debugger/live_components/traces_list.ex
```diff
@@ -218,6 +232,38 @@ defmodule LiveDebugger.LiveComponents.TracesList do
...
+        nil ->
+          assigns
+          |> assign(:callback_name, "Unknown trace")
+          |> assign(:trace_args, [])
+
+        trace ->
+          assigns
+          |> assign(:callback_name, "#{trace.function}/#{trace.arity}")
+          |> assign(:trace_args, trace.args)
+      end
```
GuzekAlan, line 250: Isn't `if` better? Also you could extract `callback_name` to separate function (it is in more than one place). But I am fine either way

GuzekAlan (approved):

PR #164, opened 2025-03-04
Task: fix streams in callback traces part. II

kraleppa:
Added lazy rendering of trace arguments in collapsible

Before this PR we were rendering the whole args preview for each trace immediately
After this PR we are rendering the args preview (content of collapsible) only when is opened for the first time

Stream is now really fast - I raised the limits and couldn't observe any lag while stopping the tracing even with the following limits:

```
  @stream_limit 512
  @traces_number 64
  @period_ms 1000
```

### DIFF
--- a/lib/live_debugger/live_components/traces_list.ex
+++ b/lib/live_debugger/live_components/traces_list.ex
@@ -159,7 +159,7 @@ defmodule LiveDebugger.LiveComponents.TracesList do
   end
 
   @impl true
-  def handle_event("toggle-collapsible", %{"data" => string_trace_id}, socket) do
+  def handle_event("toggle-collapsible", %{"trace-id" => string_trace_id}, socket) do
     trace_id = String.to_integer(string_trace_id)
 
     socket.assigns.ets_table_id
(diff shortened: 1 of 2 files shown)


lib/live_debugger/structs/trace_display.ex
```diff
@@ -0,0 +1,29 @@
...
+
+  alias LiveDebugger.Structs.Trace
+
+  defstruct [:id, :trace, :render_body?, :counter]
+
+  @type t() :: %__MODULE__{
+          id: integer(),
+          trace: Trace.t(),
+          render_body?: boolean(),
+          counter: integer() | nil
```
GuzekAlan, line 15: `non_neg_integer() | nil` ?

lib/live_debugger/live_components/traces_list.ex
```diff
@@ -148,6 +158,28 @@ defmodule LiveDebugger.LiveComponents.TracesList do
...
+        socket
+
+      trace ->
+        socket
+        |> stream_insert(
+          :existing_traces,
+          TraceDisplay.from_historical_trace(trace),
+          at: abs(trace.id),
+          limit: @stream_limit
+        )
```
GuzekAlan, line 178: Now it always inserts trace when toggled. Maybe some check to do it only once, what do you think?

kraleppa: Right, we can easily use 'render_body?' flag for that

GuzekAlan (approved):

PR #174, opened 2025-03-06
Feature: Dev tools extension

hhubert6:


GuzekAlan (commented):
Looks really nice. Good Job! Have you tested it on some application like LiveBook or LiveBeats (not the `dev`)?

assets/js/client.js
GuzekAlan, line 1: Also I don't know if there is a github action for this but could you compile assets for this?

GuzekAlan (commented):
LGTM! We need to wait for proper logo

assets/js/client.js
kraleppa, line 1: This will have to be adjusted to the new routing system

devtools/hook.js
kraleppa, line 1: Is it needed?

devtools/images/icon-128.png
kraleppa, line 1: The logo has been already choosen so you can change it

devtools/manifest.json
```diff
@@ -0,0 +1,22 @@
+{
+  "manifest_version": 3,
+  "name": "LiveDebugger DevTools",
+  "version": "0.0.0.1",
```
kraleppa, line 4:
```suggestion
  "version": "0.1.0",
```

kraleppa (commented):
It's really good, but:
- It'll have to be adjusted to new routing system after @GuzekAlan will merge #186 
- Logo has to be changed
- We need to come up with the way of publishing extension to chrome marketplace
- We need to create a feature flag to disable it for now (maybe we could just remove the plug and add a comment)

Everything should besides publishing should be handled in this PR I think

kraleppa (commented):
When there are two LiveViews opened with the same module extension and we refresh LiveDebugger cannot find active LiveView associated with given window.

The reason for this is quite simple - rediscovery mechanism works this way. However, in case chrome extension we have an access to `socket_id` of given LiveView so theoretically standard rediscovery mechanism is not needed. 

When session_id/socket_id changes in the browser we could reroute to the corresponding URL by updating it in iframe
https://github.com/user-attachments/assets/bc16fd8a-fa0e-4313-bdd3-6b469be91e0b

kraleppa (approved):
LGTM! Let's merge it 🎉

GuzekAlan (approved):
🎉

PR #178, opened 2025-03-07
Bug: Add missing font

GuzekAlan:


GuzekAlan:
In tree there was no difference between `font-normal` and `font-medium`. Adding missing font solved the issue

hhubert6 (approved):

PR #185, opened 2025-03-17
Refactor: make sections non-collapsible

GuzekAlan:


hhubert6 (approved):
LGTM

lib/live_debugger/components.ex
```diff
@@ -149,6 +149,10 @@ defmodule LiveDebugger.Components do
     """
   end
 
+  @doc """
+  Renders a collapsible section which doesn't collapse on desktop view.
```
kraleppa, line 153: Why we don't want sections to collapse on desktop?

GuzekAlan: TLDR; because we didn't want to.
In early designs of our app we've added collapsible sections when we had every section one on top of other so that you didn't have to scroll through assigns to see callbacks. It doesn't make sense to make it collapsible on big desktop view (`lg` screen). Now since assigns are already collapsible by being an elixir map we might not need to make collapsible sections at all (it would make things simpler). I think it is only the UX/UI decision whether to keep it or not

kraleppa: Ok, thanks for clarification

lib/live_debugger/components.ex
```diff
@@ -160,37 +164,34 @@ defmodule LiveDebugger.Components do
 
   def collapsible_section(assigns) do
     ~H"""
-    <div class={[
-      "w-full min-w-[20rem] lg:max-w-[32rem] h-max flex shadow-custom border border-secondary-200"
-      | List.wrap(@class)
-    ]}>
+    <div
```
kraleppa, line 167: Couldn't you just hide `chevron` icon and make window size checking via tailwind? I saw something like this:
https://v3.tailwindcss.com/docs/pointer-events 

This way maybe we could get rid of additional hook

kraleppa (commented):
Also, after I resize the screen the cursor is still pointer 
Screenshot 2025-03-19 at 09 18 40

lib/live_debugger/components.ex
GuzekAlan: It is how it was implemented in the first place. It required disabling pointer events but children should be able to have them (when we have buttons in label) but there was situation when we had 2 buttons when I clicked in space between them then the section collapsed on desktop

GuzekAlan:
> Also, after I resize the screen the cursor is still pointer
@kraleppa I cannot reproduce it. Maybe it was some uncompiled css? Can you double check in free time?

kraleppa:
> I cannot reproduce it. Maybe it was some uncompiled css? Can you double check in free time?

It seems so - it's working fine now

lib/live_debugger/components.ex
kraleppa: Not sure if I understand correctly, but if problem is that children have to have pointers events turned on couldn't we handle it this like this?

```
<div id="parent-that-needs-disabling-events" class="md:pointer-events-none">
  <div id="wrapper-for-children" class="pointer-events-auto">
    <%!--  children --%>
  </div>
</div>
```

Why am I even writing about this stuff - IMO defining screensizes once again, defining a special hook that handle each resize event seems like a lot for quite a simple effect

GuzekAlan: When we have 2 buttons next to each other in children which allows pointer events then the space between them is clickable. We could fix this by adding it to button classes (and not the wrapper) which feels weird to me. But there is also issue for non-pointer events (this is selectable by tab and can be collapsed by pressing space). But I agree that maybe hook and defining screen sizes may be a lot. As I mentioned in person do we need it (and I am not trying to adjust UI to solution but rather curious if we need to collapse these sections since assets are already collapsible by being a map)?

EDIT: This hook also ensures that this is opened when full screen. I tested `pointer-events` solution and we can collapse it on small view and when on desktop there is no button to show section again.

kraleppa: > As I mentioned in person do we need it (and I am not trying to adjust UI to solution but rather curious if we need to collapse these sections since assets are already collapsible by being a map)?

Tbh I think we don't... Info panel and assets have maximum height, and traces are at the bottom of the page so they're scorllable

kraleppa (approved):

PR #186, opened 2025-03-18
Refactor: new page routing

GuzekAlan:


lib/live_debugger/utils/url.ex
kraleppa, line 1: Maybe it should be in `live_helpers`?

lib/live_debugger/live_helpers/routes.ex
```diff
@@ -0,0 +1,33 @@
...
+  @moduledoc """
+  Helper module to generate url routes for the LiveDebugger application.
+  """
+
+  use Phoenix.VerifiedRoutes, endpoint: LiveDebugger.Endpoint, router: LiveDebugger.Router
+
+  alias LiveDebugger.Utils.Parsers
+
+  @spec live_views_dashboard() :: String.t()
+  def live_views_dashboard() do
```
kraleppa, line 11: I think it is not used anywhere 🧐

lib/live_debugger/live_views/channel_dashboard_live.ex
```diff
@@ -187,25 +197,65 @@ defmodule LiveDebugger.LiveViews.ChannelDashboard do
...
+      case {transport_pid, LiveViewDiscoveryService.lv_process(socket_id, transport_pid)} do
+        {_, nil} ->
+          nil
+
+        {nil, lv_process} ->
+          {:patch_transport_pid, lv_process}
+
+        {_, lv_process} ->
+          {:noop, lv_process}
+      end
```
kraleppa, line 242: Couldn't this function return only process struct or nil? In handle async you should be able to check if transport pid is nil. Also, coulnd't we just always trigger patching URL even when nothing has changed? Will it reload the page?

lib/live_debugger/live_views/live_views_dashboard_live.ex
```diff
@@ -107,6 +112,13 @@ defmodule LiveDebugger.LiveViews.SessionsDashboard do
     |> noreply()
   end
 
+  defp event_values_map(lv_process) do
+    %{
+      "phx-value-socket_id" => lv_process.socket_id,
+      "phx-value-transport_pid" => Parsers.pid_to_string(lv_process.transport_pid)
```
kraleppa, line 118:
```suggestion
      "phx-value-socket-id" => lv_process.socket_id,
      "phx-value-transport-pid" => Parsers.pid_to_string(lv_process.transport_pid)
```

lib/live_debugger/router.ex
```diff
@@ -11,13 +11,11 @@ defmodule LiveDebugger.Router do
...
     pipe_through([:dbg_browser])
 
-    import Phoenix.LiveView.Router
-
-    live("/", LiveDebugger.LiveViews.SessionsDashboard)
-    live("/:socket_id", LiveDebugger.LiveViews.ChannelDashboard)
-    live("/:socket_id/:node_id", LiveDebugger.LiveViews.ChannelDashboard)
+    live("/", LiveViews.LiveViewsDashboardLive, :index)
+    live("/transport_pid/:socket_id", LiveViews.ChannelDashboardLive, :patch)
+    live("/:transport_pid/:socket_id", LiveViews.ChannelDashboardLive, :show)
```
kraleppa, line 19: Why do we name actions in these? I don't see usage of `@live_action` anywhere

kraleppa (commented):
Great job! Some suggestions:

lib/live_debugger/router.ex
GuzekAlan: Originally I've used it to tell when to patch url but it changed so these can be omitted

lib/live_debugger/utils/url.ex
GuzekAlan: I thought of this as rather general utility. Routes were something more connected to LiveView and it uses this module. But I can move it since both places feel right to me

kraleppa: Ok I guess it can stay here

kraleppa (approved):

PR #187, opened 2025-03-18
Feature: components highlighting

hhubert6:


priv/static/app.css
GuzekAlan, line 1: you can revert it to not have conflicts

lib/live_debugger/components.ex
```diff
@@ -577,6 +577,26 @@ defmodule LiveDebugger.Components do
...
+
+  def toggle_switch(assigns) do
+    ~H"""
+    <label class="inline-flex items-center cursor-pointer pr-6 py-3">
+      <span class="text-xs font-normal text-primary-text mx-2">
+        <%= @label %>
+      </span>
+      <input id="highlight-switch" type="checkbox" class="sr-only peer" checked={@checked} {@rest} />
+      <div class="relative w-9 h-5 bg-surface-2-bg-hover peer-focus:outline-none peer-focus:ring-2 peer-focus:ring-button-primary-bg-hover rounded-full peer peer-checked:after:translate-x-full rtl:peer-checked:after:-translate-x-full after:content-[''] after:absolute after:top-[2px] after:start-[2px] after:bg-surface-0-bg after:rounded-full after:h-4 after:w-4 after:transition-all peer-checked:bg-button-primary-bg ">
+      </div>
```
GuzekAlan, line 595: This is massive. If this is copied, can you add link to where did you copy it from?

lib/live_debugger/components/tree.ex
```diff
@@ -30,7 +31,17 @@ defmodule LiveDebugger.Components.Tree do
...
+      <div class="flex items-center justify-between">
+        <div class="shrink-0 font-medium text-secondary-text px-6 py-3"><%= @title %></div>
+        <%= if Application.get_env(:live_debugger, :browser_features?) do %>
+          <.toggle_switch
+            label="Highlight"
+            checked={@highlight?}
+            phx-target={@event_target}
+            phx-click="toggle-highlight"
+          />
+        <% end %>
```
GuzekAlan, line 43: I don't know if we want to add it in app before `0.2.0`. Maybe some feature flag which will be set to off like in `toggle dark mode`? @kraleppa what do you think?

GuzekAlan (commented):
I don't know if this is a bug or a feature but when you hover over components tree and have focused highlighting switch you can disable highlighting with component selected:

image

lib/live_debugger/components/tree.ex
kraleppa: yup, feature flag is needed here

GuzekAlan (approved):

assets/js/client.js
kraleppa, line 1: Since we are bundling `client.js` couldn't we perhaps split each functionality to the separate file?

assets/js/client.js
```diff
@@ -125,6 +125,132 @@ document.addEventListener('DOMContentLoaded', function () {
...
+    let highlightElement = document.getElementById(highlightElementID);
+
+    if (highlightElement) {
+      highlightElement.remove();
+      if (
+        msg.detail.val === undefined ||
+        highlightElement.dataset.val === msg.detail.val
+      ) {
+        return;
+      }
```
kraleppa, line 173: Maybe like this? If you don't like it, feel free to ignore
```suggestion
      const cond = msg.detail.val === undefined || highlightElement.dataset.val === msg.detail.val
      if (cond) return;
```

assets/js/client.js
```diff
@@ -125,6 +125,132 @@ document.addEventListener('DOMContentLoaded', function () {
...
+        highlightElementID
+      );
+      document.body.appendChild(highlightElement);
+    }
+  });
+
+  window.addEventListener('resize', () => {
+    const highlight = document.getElementById(highlightElementID);
+    if (highlight) {
+      const activeElement = document.querySelector(
```
kraleppa, line 193: Why do you declare it here as `const` - it is declared in line 161 via `let`

assets/js/hooks/highlight.js
```diff
@@ -0,0 +1,28 @@
...
+
+    this.pushHighlight = (e) => {
+      if (highlightSwitch.checked) {
+        params = {
+          search_attribute:
+            e.target.attributes['phx-value-search_attribute'].value,
+          search_value: e.target.attributes['phx-value-search_value'].value,
+        };
+        this.pushEventTo('#sidebar', 'highlight', params);
+      }
```
kraleppa, line 14:
```suggestion
      if (highlightSwitch.checked) {
        const attr = e.target.attributes;
        params = {
          search_attribute: attributes['phx-value-search_attribute'].value,
          search_value: attributes['phx-value-search_value'].value,
        };
        this.pushEventTo('#sidebar', 'highlight', params);
      }
```

lib/live_debugger/components/tree.ex
```diff
@@ -139,9 +150,13 @@ defmodule LiveDebugger.Components.Tree do
 
     ~H"""
     <button
+      id={"tree_node_button_" <> @node.parsed_id}
       phx-click="select_node"
       phx-value-node_id={@node.parsed_id}
       phx-target={@event_target}
+      phx-value-search_attribute={get_search_attribute(@node)}
+      phx-value-search_value={get_search_value(@node)}
```
kraleppa, line 158:
```suggestion
      phx-value-search-attribute={get_search_attribute(@node)}
      phx-value-search-value={get_search_value(@node)}
```

kraleppa (approved):

PR #191, opened 2025-03-19
Feature: Improve display of structs

hhubert6:


lib/live_debugger/components/elixir_display.ex
```diff
@@ -32,15 +32,17 @@ defmodule LiveDebugger.Components.ElixirDisplay do
         id={@id <> "collapsible"}
         open={@expanded?}
         icon="icon-chevron-right"
-        class="[&>summary_.hide-on-opened]:open:hidden"
+        class="[&>summary_.hide-on-opened]:open:hidden [&>summary_.show-on-opened]:open:flex"
```
GuzekAlan, line 35: I found that this is duplicated in `collapsible/1`. Can you please delete it from here and move there?

GuzekAlan (approved):

PR #194, opened 2025-03-20
Task: fix code reloading

hhubert6:


lib/live_debugger.ex
```diff
@@ -27,7 +27,13 @@ defmodule LiveDebugger do
...
         secret_key_base: Keyword.get(config, :secret_key_base, @default_secret_key_base),
         live_view: [signing_salt: Keyword.get(config, :signing_salt, @default_signing_salt)],
-        adapter: Keyword.get(config, :adapter, default_adapter)
+        adapter: Keyword.get(config, :adapter, default_adapter),
+        live_reload: [
+          patterns: [
+            ~r"priv/static/.*(js|css|svg)$",
+            ~r"priv/static/dev/.*(js|css|svg)$"
+          ]
+        ]
```
GuzekAlan, line 36: Isn't it code which will be executed on prod? We want `live_reload` on our `dev` only

config/config.exs
```diff
@@ -14,7 +14,7 @@ if config_env() == :dev do
     ],
     dev_build: [
       args:
-        ~w(js/app.js js/client.js --bundle --minify --sourcemap=external --target=es2020 --outdir=../priv/static/dev),
+        ~w(js/app.js js/client.js --bundle --sourcemap=external --target=es2020 --outdir=../priv/static/dev),
```
kraleppa, line 17: Why did you delete the minify?

lib/live_debugger.ex
kraleppa: It'll this cannot be merged

config/config.exs
hhubert6: Do we really need minify on dev? It makes building assets 3x longer which is annoying when you're making changes to assets and want fast reload.

kraleppa (approved):

PR #200, opened 2025-03-21
Bug: fix mismatch in LiveView deps

GuzekAlan:


kraleppa:
Screenshot 2025-03-21 at 11 57 11 💀

GuzekAlan:
> Screenshot 2025-03-21 at 11 57 11 💀

I am currently using this branch to test and draft PR helps me with checking diffs. Ignore this draft PR cause some weird things might appear :)

README.md
```diff
@@ -22,6 +22,18 @@ Add `live_debugger` to your list of dependencies in `mix.exs`:
 
 After you start your application LiveDebugger will be running at a default port `http://localhost:4007`.
 
+### Assets rebuilding
```
kraleppa, line 25: Tbh I think it should not be optional. It should be a compulsory step when downloading a LiveView debugger. We should also consider code generation I mentioned IRL. I think we'll take care of it after 0.2.0

mix.exs
```diff
@@ -48,7 +53,7 @@ defmodule LiveDebugger.MixProject do
       {:phoenix_live_view, "~> 0.20 or ~> 1.0"},
       {:bandit, "~> 1.6", only: :dev},
       {:phoenix_live_reload, "~> 1.5", only: :dev},
-      {:esbuild, "~> 0.7", only: :dev},
+      {:esbuild, "~> 0.7", only: [:dev, :prod]},
```
kraleppa, line 56: If we have to depend on `esbuild` we'll have to move it just below LiveView, and probably adjust accepted `esbuild` versions. This should work on as many `esbuild` as possible

lib/mix/tasks/live_debugger.assets.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Mix.Tasks.LiveDebugger.Assets do
+  @moduledoc """
+  Task for generating assets for LiveDebugger
+  This task is maent to be used in development only by users of `live_debugger` library.
```
kraleppa, line 4:
```suggestion
  This task is meant to be used in development only by users of `live_debugger` library.
```

README.md
kraleppa: We'll have to also somehow make sure that the user does not have to rebuild assets after each download of the project. Maybe we should keep these files somewhere else

assets/js/setup.js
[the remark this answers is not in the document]
GuzekAlan: It is because we want to compile our files (hooks, etc) to js but don't compile `phoenix` and `phoenix_live_view`. Then we can combine them into final `app.js`

PR #204, opened 2025-03-26
Bug: Fixed mismatch in LiveView browser and server deps

kraleppa:


lib/live_debugger.ex
```diff
@@ -27,8 +27,7 @@ defmodule LiveDebugger do
         http: [ip: ip, port: port],
         secret_key_base: Keyword.get(config, :secret_key_base, @default_secret_key_base),
         live_view: [signing_salt: Keyword.get(config, :signing_salt, @default_signing_salt)],
-        adapter: Keyword.get(config, :adapter, default_adapter),
-        live_reload: Keyword.get(config, :live_reload, [])
```
hhubert6, line 31: Why exactly did you removed `live_reload`? It is needed for live reloading assets on dev.

kraleppa: uh my mistake - thanks, good catch

lib/live_debugger/layout.ex
```diff
@@ -24,12 +24,44 @@ defmodule LiveDebugger.Layout do
...
       </head>
       <body class="theme-light text-primary-text bg-main-bg text-xs font-normal">
+        <script src="/assets/phoenix/phoenix.js">
+        </script>
+        <script src="/assets/phoenix_live_view/phoenix_live_view.js">
+        </script>
+        <script src="/assets/live_debugger/hooks.js">
+        </script>
+        <script>
+          function saveDialogAndDetailsState() {
```
GuzekAlan, line 39: This function could be moved to hooks.js

GuzekAlan (approved):

PR #210, opened 2025-03-27
Task: Update UI in channel dashboard for nested LiveViews

GuzekAlan:


lib/live_debugger/structs/lv_process.ex
```diff
@@ -64,4 +67,11 @@ defmodule LiveDebugger.Structs.LvProcess do
         nil
     end
   end
+
+  @doc """
+  Returns the parent LvProcess of the given `lv_process`.
+  """
+  @spec parent(lv_process :: t()) :: t() | nil
+  def parent(%__MODULE__{parent_pid: nil}), do: nil
+  def parent(%__MODULE__{parent_pid: parent_pid}), do: new(parent_pid)
```
GuzekAlan, line 76: not sure if this should be here or in LiveViewDiscorveryService. It only expands one of the field of `LvProcess` so IMO it could be here. What do you think?

kraleppa: imo it's ok

lib/live_debugger/live_components/sidebar.ex
```diff
@@ -166,26 +181,103 @@ defmodule LiveDebugger.LiveComponents.Sidebar do
...
+        <div :if={parent_lv_process} class="w-full flex flex-col">
+          <span class="font-medium">Parent LiveView Process</span>
+          <.live_view_link lv_process={parent_lv_process} id="parent_live_view_link" icon?={false} />
+        </div>
+      </.async_result>
+    </div>
+    """
+  end
+
+  attr(:nested_lv_processes, :any, required: true)
```
kraleppa, line 214:
```suggestion
  attr(:nested_lv_processes, :list, required: true)
```

lib/live_debugger/live_components/sidebar.ex
```diff
@@ -166,26 +181,103 @@ defmodule LiveDebugger.LiveComponents.Sidebar do
...
+            ]
+          }
+          class="w-full flex flex-col"
+        >
+          <span class="font-medium"><%= text %></span>
+          <span><%= value %></span>
+        </div>
+        <div :if={parent_lv_process} class="w-full flex flex-col">
+          <span class="font-medium">Parent LiveView Process</span>
+          <.live_view_link lv_process={parent_lv_process} id="parent_live_view_link" icon?={false} />
```
kraleppa, line 207:
```suggestion
          <.live_view_link lv_process={parent_lv_process} id="parent-live-view-link" icon?={false} />
```

lib/live_debugger/live_components/sidebar.ex
```diff
@@ -166,26 +181,103 @@ defmodule LiveDebugger.LiveComponents.Sidebar do
...
+        </div>
+      </.async_result>
+    </div>
+    """
+  end
+
+  attr(:nested_lv_processes, :any, required: true)
+
+  defp nested_live_views_links(assigns) do
```
kraleppa, line 216: If we load additional data specificly for `nested_live_views_links` component then maybe we should refactor it to a separate `LiveComponent`? 🤔  

This way loading this data could be encapsulated inside the component that uses them

GuzekAlan: I feel like if this is a `LiveComponent` then `ComponentTree` should also be one. It will require passing information to them after getting a trace but it should be more clean. What do you think? I wouldn't change `ComponentsTree` in this PR but rather after changing to nested LiveView (Sidebar)

kraleppa: Yup it might make sense to change component tree too. I'll refactor it when changing sidebar to LiveView

kraleppa (approved):

PR #221, opened 2025-03-28
Fix: browser features not working with embedded live views in controller

hhubert6:
When debugged application has a controller with embedded live view, browser features are not working. It's because there is no element on the page with attribute `data-phx-main`.

### DIFF
--- a/assets/js/client.js
+++ b/assets/js/client.js
@@ -21,9 +21,9 @@ window.getLiveDebuggerURL = function () {
     .src.replace('/assets/live_debugger/client.js', '');
 
   const session_id = getSessionId();
-  const session_path = session_id ? `/transport_pid/${session_id}` : '';
+  const session_path = session_id ? `transport_pid/${session_id}` : '';
 
-  return `${baseURL}${session_path}`;
+  return `${baseURL}/${session_path}`;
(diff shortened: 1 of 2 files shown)


GuzekAlan (approved):

assets/js/client.js
```diff
@@ -2,20 +2,34 @@
...
-  return document
+  const baseURL = document
     .getElementById('live-debugger-scripts')
     .src.replace('/assets/live_debugger/client.js', '');
-};
 
-window.getSessionId = function () {
-  return document.querySelector('[data-phx-main]').id;
+  const session_id = getSessionId();
+  const session_path = session_id ? `/transport_pid/${session_id}` : '';
```
kraleppa, line 24:
```suggestion
  const session_path = session_id ? `transport_pid/${session_id}` : '';
```

assets/js/client.js
```diff
@@ -2,20 +2,34 @@
...
+
+  return `${baseURL}${session_path}`;
```
kraleppa, line 26:
```suggestion
  return `${baseURL}/${session_path}`;
```

kraleppa (approved):

PR #227, opened 2025-03-31
Feature: Introduce flash messages mechanism

GuzekAlan:


GuzekAlan:
@kraleppa which messages should we add? I've only added one error to be displayed for now

kraleppa:
Imo in case of dark mode background should be dark

Screenshot 2025-03-31 at 15 53 47

Similarly to

Screenshot 2025-03-31 at 15 55 53

GuzekAlan:
@kraleppa 

> Also, I'm not 100% if having them in right-upper corner is a good idea. Maybe lower left could be better?

I think it might be good idea because it is least used corner

> Imo in case of dark mode background should be dark

I've tried using it but it blends with background. Maybe adding red border will be okay but I feel it still won't be differentiable. I've used tailwind colors for now because we don't have colors for sucessful messages (but we can not add them at the moment since we rather use it for errors instead of confirmations). What do you think?

lib/live_debugger/flash.ex
```diff
@@ -0,0 +1,35 @@
...
+  @doc """
+  Function to put flash inside nested LiveViews/LiveComponents.
+  If used in nested LiveView use parent's pid.
+  """
+  @spec put_flash!(
+          socket :: Phoenix.LiveView.Socket.t(),
+          message :: String.t()
+        ) ::
+          Phoenix.LiveView.Socket.t()
+  def put_flash!(pid \\ self(), socket, message) do
```
kraleppa, line 24: 1. Why does it use `!` convention? It doesn't raise any errors
2. To distinct it from sthe tandard `put_flash`, you could name it `push_flash`

lib/live_debugger/components.ex
```diff
@@ -123,6 +123,61 @@ defmodule LiveDebugger.Components do
...
+    >
+      <div class="flex gap-3 items-center">
+        <.icon name="icon-x-circle" class="text-red-500" />
+        <p>
+          <%= @message %>
+        </p>
+      </div>
+      <button
+        phx-click={
+          JS.push("lv:clear-flash", value: %{key: :error})
```
kraleppa, line 166:
```suggestion
          "lv:clear-flash"
          |> JS.push(value: %{key: :error})
```

kraleppa (approved):
Good job 🥇 Just some nitpicks

lib/live_debugger/flash.ex
GuzekAlan: since we use `Flash` module we could alias it and name function as `put` (Flash.put/2)`, what do you think ?

GuzekAlan:
@hhubert6 

> When I click multiple times fast, it starts to weirdly blink

It shouldn't blink now

> Also, if there is a button to dismiss the flash message, wouldn't it be better if it stayed on the screen for longer?

I found solution which gave 5s on screen. It seemed a little long for me but maybe because of the small amount of text. We could check the size of the content (e.g. word count) and show message based on that but to me arbitrary setting one value looks better. How much longer do you think it should stay?

hhubert6 (approved):
@GuzekAlan 
I think it was related to this issue with blinking that it disappeared too quickly sometimes. 5s is fine to me.

PR #231, opened 2025-04-01
Enhancement: implement new routing mechanism

kraleppa:


lib/live_debugger/structs/trace.ex
```diff
@@ -31,17 +44,40 @@ defmodule LiveDebugger.Structs.Trace do
...
       id: id,
       module: module,
       function: function,
       arity: length(args),
       args: args,
+      socket_id: socket_id,
+      transport_pid: transport_pid,
       pid: pid,
       cid: cid,
       timestamp: :os.system_time(:microsecond)
```
GuzekAlan, line 83: Cannot this all be one function with keyword args (`cid`, `socket_id`, `transport_pid`) and it will take from args when none is provided?

lib/live_debugger/utils/pubsub.ex
```diff
@@ -2,32 +2,91 @@ defmodule LiveDebugger.Utils.PubSub do
...
   end
 
-  # FYI this topic is temporary, it will be replaced after the new tracing system is implemented
-  @spec session_trace_topic(lv_process :: LvProcess.t()) :: String.t()
-  def session_trace_topic(lv_process) do
-    "lvdbg/#{inspect(lv_process.transport_pid)}/new_trace"
+  @spec trace_topic(String.t(), pid()) :: String.t()
+  def trace_topic(socket_id, transport_pid) do
+    "#{inspect(transport_pid)}/#{socket_id}/*/*"
   end
```
GuzekAlan, line 91: How about returning multiple topics. I mean we pass arguments which are necessary and it gives us all topics where these parameters are combined? E.g. for `trace_topic/3` returns its result and `trace_topic/2` result. Then we can have cascading topics and only need to call one (most specific) to get all (generic) topics?

lib/live_debugger/live_helpers/tracing_helper.ex
```diff
@@ -81,6 +82,10 @@ defmodule LiveDebugger.LiveHelpers.TracingHelper do
       fuse: %{count: 0, start_time: now()}
     }
 
+    if Phoenix.LiveView.connected?(socket) do
+      PubSubUtils.subscribe(socket.assigns.trace_topic)
+    end
+
```
GuzekAlan, line 88: Maybe use some `Map.get` for when there will be no `trace_topic` assign? Or add some information in moduledoc

GuzekAlan (commented):
Nice 🦊

lib/live_debugger/structs/trace.ex
kraleppa: Good idea 👍

lib/live_debugger/utils/pubsub.ex
kraleppa: Tbh I wouldn't get attached to these topics because they'll probably changed during implementation of filtering

lib/live_debugger/live_helpers/tracing_helper.ex
kraleppa: Another good idea 👌

hhubert6 (commented):
When running tests this error is logged

```
10:14:42.374 [error] GenServer LiveDebugger.GenServers.CallbackTracingServer terminating
** (Mox.UnexpectedCallError) no expectation defined for LiveDebugger.MockModuleService.all/0 in process #PID<0.312.0> with args []
    (mox 1.2.0) lib/mox.ex:903: Mox.__dispatch__/4
    (live_debugger 0.1.4) lib/live_debugger/services/module_discovery_service.ex:13: LiveDebugger.Services.ModuleDiscoveryService.all_modules/0
    (live_debugger 0.1.4) lib/live_debugger/gen_servers/callback_tracing_server.ex:33: LiveDebugger.GenServers.CallbackTracingServer.handle_continue/2
    (stdlib 6.1.2) gen_server.erl:2335: :gen_server.try_handle_continue/3
    (stdlib 6.1.2) gen_server.erl:2244: :gen_server.loop/7
    (stdlib 6.1.2) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
Last message: {:continue, :setup_tracing}
State: []
```

GuzekAlan (commented):
Apart from error in tests LGTM!

mix.exs
```diff
@@ -16,7 +16,8 @@ defmodule LiveDebugger.MixProject do
       name: "LiveDebugger",
       source_url: "https://github.com/software-mansion/live-debugger",
       description: "Tool for debugging LiveView applications",
-      docs: docs()
+      docs: docs(),
+      preferred_cli_env: [e2e: :test]
```
GuzekAlan, line 20: Shouldn't this be configured in `cli/0` function ?
https://hexdocs.pm/mix/1.16.1/Mix.Project.html#module-cli-configuration

lib/live_debugger.ex
```diff
@@ -61,4 +61,9 @@ defmodule LiveDebugger do
       {:error, _} -> Phoenix.Endpoint.Cowboy2Adapter
     end
   end
+
+  defp unit_test?() do
+    Application.get_env(:live_debugger, :test_mode?, false) and
+      not Application.get_env(:live_debugger, :e2e?, false)
+  end
```
GuzekAlan, line 68: IMO it looks like it could be one of the `LiveDebugger.Env` methods but I am fine either way

GuzekAlan (approved):

PR #247, opened 2025-04-08
Task: attach tracing filters to logic

kraleppa:


hhubert6 (commented):
When filters and trace arguments are opened they overlap each other.

image

lib/live_debugger/utils/pubsub.ex
```diff
@@ -32,26 +24,19 @@ defmodule LiveDebugger.Utils.PubSub do
     Phoenix.PubSub.subscribe(LiveDebugger.PubSub, topic)
   end
 
+  @spec unsubscribe(topics :: [String.t()]) :: :ok
+  def unsubscribe(topics) when is_list(topics) do
+    topics
+    |> Enum.each(&unsubscribe(&1))
+
+    :ok
+  end
```
GuzekAlan, line 33: We really omit any error handling here. How about raising on 1 element subscribe/unsubscribe when it returns error? Wouldn't it make it easier to debug PubSub in the future?

lib/live_debugger/utils/pubsub.ex
```diff
@@ -70,23 +55,34 @@ defmodule LiveDebugger.Utils.PubSub do
...
+  @doc """
+  It stands for transport_pid/socket_id/node_id/function.
 
-  @spec trace_topic(String.t(), pid(), atom()) :: String.t()
-  def trace_topic(socket_id, transport_pid, fun) when is_atom(fun) do
-    "#{inspect(transport_pid)}/#{socket_id}/*/#{inspect(fun)}"
+  It gives you traces of given callback in given node in given LiveView
+  Used to update assigns based on render callback and for filtering traces
+  """
+  @spec tsnf_topic(
```
GuzekAlan, line 64: Docs are explaining what these names mean but at first glance they look weird. Why not use something like `format: :ts_f` as opts and name it same way?

lib/live_debugger/services/trace_service.ex
```diff
@@ -83,21 +99,15 @@ defmodule LiveDebugger.Services.TraceService do
     |> :ets.match_object(matcher, limit)
     |> case do
       {entries, _cont} ->
-        Enum.map(entries, &elem(&1, 1))
+        entries
+        |> Enum.map(&elem(&1, 1))
+        |> filter_by_functions(functions)
```
GuzekAlan, line 104: Correct me if I am wrong but doesn't it mean that we fetch 100 recrds, filter so that 0 is left and show nothing when limit is 100? Can the filtering by function be done inside `:ets` lookup

lib/live_debugger/utils/pubsub.ex
kraleppa: I did it because I wanted: 
1. Made it easy to search via `cmd + shift + f` - thanks to this you can easily find where given topic was used
2. Erlang sometimes uses this convention e.g. `mfa` (Module Function Arity)

lib/live_debugger/live_helpers/tracing_helper.ex
```diff
@@ -30,6 +29,19 @@ defmodule LiveDebugger.LiveHelpers.TracingHelper do
...
+    socket
+    |> get_topics()
+    |> PubSubUtils.unsubscribe()
+
+    socket
+    |> get_active_topics()
+    |> PubSubUtils.subscribe()
+
+    socket
+  end
```
hhubert6, line 43: I think this function is not needed because when you apply filters you already stopped tracing (and unsubscribed to topics). After applying filters you will start tracing which will subscribe to all needed topics.

Currently you're subscribing to topics twice.

kraleppa: you're right - deleted 🙏

GuzekAlan (approved):

PR #259, opened 2025-04-10
Fix: Add missing flash message

hhubert6:


lib/live_debugger/live_views/traces_live.ex
```diff
@@ -177,8 +180,11 @@ defmodule LiveDebugger.LiveViews.TracesLive do
         |> stream_insert(:existing_traces, trace_display, at: 0, limit: @stream_limit)
         |> assign(:traces_empty?, false)
 
+      {:stopped, socket} ->
+        socket.assigns.root_pid
+        |> push_flash(socket, "Trace limit exceeded")
```
GuzekAlan, line 185: Could you add "Callback tracer stopped!" before?

GuzekAlan (approved):
As discussed in person could you please format the other flash message (to add italic instead of  \` \`?

GuzekAlan (approved):
🪿

PR #260, opened 2025-04-11
Task: Implement caching mechanism

GuzekAlan:


GuzekAlan:
I've tested it on livebook and it worked rather smoothly. I've added feature flag to choose weather to cache state or not (might be connected with `:dead_view` mode in the future)

kraleppa:
Closing since it'll have to be adjusted to the new version of callback tracing

PR #263, opened 2025-04-14
Feature: add pagination in history mode

kraleppa:


lib/live_debugger/services/trace_service.ex
```diff
@@ -67,29 +67,43 @@ defmodule LiveDebugger.Services.TraceService do
...
+      cont
+      |> :ets.select()
+      |> case do
+        {entries, new_cont} ->
+          {Enum.map(entries, &elem(&1, 1)), new_cont}
+
+        :"$end_of_table" ->
+          :"$end_of_table"
+      end
     end
```
GuzekAlan, line 107: maybe split into separate functions and call functions in if/else

lib/live_debugger/services/trace_service.ex
```diff
@@ -67,29 +67,43 @@ defmodule LiveDebugger.Services.TraceService do
   ## Options
     * `:node_id` - PID or CID to filter traces by
     * `:limit` - Maximum number of traces to return (default: 100)
+    * `:cont` - Used to get next page of items in the following queries
     * `:functions` - List of function names to filter traces by
   """
-  @spec existing_traces(atom(), keyword()) :: [Trace.t()]
+  @spec existing_traces(atom(), keyword()) :: {[Trace.t()], term()} | :"$end_of_table"
```
GuzekAlan, line 73: why return ets atom? We could return `:empty` or `nil` instead if we wrap ETS logic in this module

lib/live_debugger/live_views/traces_live.ex
```diff
@@ -273,173 +320,36 @@ defmodule LiveDebugger.LiveViews.TracesLive do
...
+  end
+
+  defp load_more_existing_traces(socket) do
+    ets_table_id = socket.assigns.ets_table_id
+    node_id = socket.assigns.node_id
+    cont = socket.assigns.traces_continuation
+    active_functions = get_active_functions(socket)
+
+    socket
+    |> assign(:traces_continuation, :loading)
```
GuzekAlan, line 347: Why not use `AsyncResult` and `<.async_result>` in the HEEX?

GuzekAlan (commented):
Nice! 🚀  now users will load all traces 👍

lib/live_debugger/live_views/traces_live.ex
kraleppa: We are attaching loaded data to stream - AsyncResult does not have support for streams, so we have to do it manually

lib/live_debugger/services/trace_service.ex
kraleppa: imo `:end_of_table` is more desrciptiove that `:empty` or `nil` that may suggest that the whole table is empty. But we can get rid of these weird erlang characters

lib/live_debugger/services/trace_service.ex
```diff
@@ -67,29 +67,26 @@ defmodule LiveDebugger.Services.TraceService do
...
-    node_id = Keyword.get(opts, :node_id)
-
-    if limit < 1 do
-      raise ArgumentError, "limit must be >= 1"
+    opts
+    |> Keyword.get(:cont, nil)
+    |> case do
+      nil -> existing_traces_start(table_id, opts)
+      _cont -> existing_traces_continuation(opts)
     end
```
GuzekAlan, line 80: You could also check if the `:cont` is not a `:end_of_table` just to be sure no error occurs

lib/live_debugger/live_views/traces_live.ex
```diff
@@ -120,46 +119,88 @@ defmodule LiveDebugger.LiveViews.TracesLive do
...
               <%= if wrapped_trace.id == "separator" do %>
-                <.separator id={dom_id} />
+                <Traces.separator id={dom_id} />
               <% else %>
-                <.trace id={dom_id} wrapped_trace={wrapped_trace} />
+                <Traces.trace id={dom_id} wrapped_trace={wrapped_trace} />
               <% end %>
             <% end %>
           </div>
+          <div class="flex items-center justify-center mt-4">
```
GuzekAlan, line 128: why not add feature flag check here?

GuzekAlan (approved):
👍

PR #268, opened 2025-04-15
Enhancement: responsive UI improvements

kraleppa:


lib/live_debugger/live_views/channel_dashboard_live.ex
```diff
@@ -45,7 +45,7 @@ defmodule LiveDebugger.LiveViews.ChannelDashboardLive do
           <.nav_icon
             :if={@lv_process.ok?}
             phx-click={JS.push("open-sidebar", target: "#sidebar")}
-            class="flex md:hidden"
+            class="flex xl:hidden"
```
hhubert6, line 48:
```suggestion
            class="flex lg:hidden"
```

It should be like this I think. It shows when sidebar is visible otherwise.

kraleppa: Good catch!

PR #277, opened 2025-04-17
Task: implement displaying event handling time

hhubert6:


GuzekAlan (approved):

GuzekAlan (changes requested):
image
 
 When opening callback when it is still being run we end up in infinite loop

kraleppa (commented):
Maybe some e2e tests? 👀

assets/js/hooks/trace_execution_time.js
```diff
@@ -2,16 +2,23 @@ const TraceExecutionTime = {
...
       current = Date.now() - start;
       this.el.textContent = current + ' ms';
     }, 16);
 
     this.handleEvent('stop-timer', () => {
-      clearInterval(this.intervalId);
+      if (!handled) {
+        clearInterval(this.intervalId);
+        this.el.closest('details').open = false;
+      }
```
GuzekAlan, line 16: why is this used? shouldn't there be some `handled = true` somewhere?

GuzekAlan (approved):

PR #291, opened 2025-04-23
Refactor: Components tree styling

GuzekAlan:


lib/live_debugger/live_components/live_dropdown.ex
```diff
@@ -62,7 +62,7 @@ defmodule LiveDebugger.LiveComponents.LiveDropdown do
         variant={Map.get(button_slot, :variant, "secondary")}
         size={Map.get(button_slot, :size, "sm")}
         id={@id <> "-button"}
-        phx-click={unless @open, do: "open"}
+        phx-click={if !@open, do: "open"}
```
kraleppa, line 65: Why unless is no good here?

lib/live_debugger/components.ex
```diff
@@ -103,7 +103,7 @@ defmodule LiveDebugger.Components do
   @doc """
   Collapsible element that can be toggled open and closed.
   It uses the `details` and `summary` HTML elements.
-  If you add `hide-on-open` class to element it will be hidden when collapsible is opened.
+  `hide-on-open` and `show-on-open` css classes are used to hide or show elements based on the open state of the collapsible.
```
kraleppa, line 106: Not sure if we want to go this way tbh. For example in VSCode it works exactly the same as on main

https://github.com/user-attachments/assets/28e861e9-b432-4710-bb87-2fef9c682850

lib/live_debugger/live_components/live_dropdown.ex
GuzekAlan: I tried `mix format --migrate`. `unless` is being deprecated: https://hexdocs.pm/elixir/Kernel.html#unless/2

lib/live_debugger/components.ex
GuzekAlan: I've made it so that only clicking on text or icon will select node (same with highlighting). Also I've added underline to indicate it. I feel that UX is rather ok but please tell me what do you think of this approach

kraleppa: I think it's fine this way tbh 👌

lib/live_debugger/components.ex
```diff
@@ -142,9 +142,8 @@ defmodule LiveDebugger.Components do
       {show_collapsible_assign(@open)}
     >
       <summary
-        class={[
-          "block flex items-center cursor-pointer" | List.wrap(@label_class)
-        ]}
+        id={@id <> "-summary"}
+        class={["block flex items-center cursor-pointer" | List.wrap(@label_class)]}
```
kraleppa, line 146: I think block and flex exclude each other - they are both values for `display` property

kraleppa (approved):

PR #292, opened 2025-04-23
Enhancement: Updated styling for scrollbars

kraleppa:


GuzekAlan (commented):
Looks good 🎸
Is it making default scroll as the thinner one? If yes then maybe we should add thick scroll for the ActiveLiveViews as well?

GuzekAlan (approved):

PR #294, opened 2025-04-23
Bug: Fix assigns not scrollable

GuzekAlan:


lib/live_debugger/live_views/channel_dashboard_live.ex
```diff
@@ -72,17 +72,14 @@ defmodule LiveDebugger.LiveViews.ChannelDashboardLive do
...
-            </div>
+            <StateLive.live_render
+              id="node-state-lv"
+              socket={@socket}
+              lv_process={lv_process}
+              node_id={@node_id || lv_process.pid}
+            />
             <TracesLive.live_render
               id="traces-list"
-              class="xl:w-1/2"
```
kraleppa, line 85: Imo it's better when you put `xl:w-1/2` here not inside nested LiveView. It's easier to understand and control layout of channel dashboard when classes resposible for high-level positioning of channel dashboard components are declared in one place

kraleppa (approved):

Issue #248, opened 2025-04-09
Add tracking of the data memory size loaded by each live_view


edwardcashmere:
### Problem Description

I have worked on some projects where OOMs happened a couple of times. This would be an amazing feature on top of what you already have 

### Proposed Solution

Investigate how `: observer` tracks process memory. I used it to debug this issue, and it worked very well. Or track the size of the data in the assigns.

kraleppa:
Hi @edwardcashmere, nice idea! We'll take a look at it 👌

nathanmalishev:
In this PR https://github.com/software-mansion/live-debugger/pull/312 - I have added some memory info for the LiveView and the Assigns.

PR #312, opened 2025-04-25
Added liveview memory usage + estimate of the assigns

nathanmalishev:
Added some basic memory usage for Liveview & also assigns.

Because the GC can have such an impact on the size, i've opted for two values 1 before & 1 after.

I wasn't quite sure on how to estimate the assigns size, so i've taken a stab (please note this is not just the user defined area of assigns).

Also after some initial testing I found that adding a simple poll to re-gauge memory usage after hydration, was more useful than the initial numbers that were returned.

I also didn't find much in the way of tests for testing the main dashboard here, so didn't add any tests. But please push back if they are required / for any other changes.

Related to issue: https://github.com/software-mansion/live-debugger/issues/248

image


lib/live_debugger/live_views/sidebar_live.ex
```diff
@@ -104,12 +112,49 @@ defmodule LiveDebugger.LiveViews.SidebarLive do
...
     </div>
     """
   end
 
+  @impl true
+  def handle_info(:assign_memory_usage, socket) do
+    Process.send_after(self(), :assign_memory_usage, 2000)
+
+    raw =
+      case :erlang.process_info(self(), :memory) do
```
kraleppa, line 129: `self()` in this context returns a PID of `SidebarLive` LiveView - not debugged LiveView process

lib/live_debugger/live_views/sidebar_live.ex
```diff
@@ -104,12 +112,49 @@ defmodule LiveDebugger.LiveViews.SidebarLive do
...
+
+    post_gc =
+      case :erlang.process_info(self(), :memory) do
+        {:memory, bytes} -> bytes
+        _ -> 0
+      end
+
+    assigns_size =
+      try do
+        words = :erts_debug.size(socket.assigns)
```
kraleppa, line 144: As above - you are calculating here size of `SidebarLive` assigns

kraleppa (changes requested):

lib/live_debugger/live_views/sidebar_live.ex
nathanmalishev: My bad @kraleppa wasn't thinking - targeting the `lv_process.root_pid` which i believe will be the correct one now.

lib/live_debugger/live_views/sidebar_live.ex
nathanmalishev: Yep - thanks. Followed other example of using the ChannelService to pull in the node and then calculate the size on that assigns. @kraleppa

nathanmalishev:
Made changes to fetch correct PID / correct Assigns. Now with the correct data the pre garbage collection values stabilize with the post garbage collection values as expected (but you can see the initial spike which may be helpful). Maybe as a further enhancement a simple mem graph could be implemented.

lib/live_debugger/live_views/sidebar_live.ex
```diff
@@ -104,12 +114,55 @@ defmodule LiveDebugger.LiveViews.SidebarLive do
...
     """
   end
 
+  @impl true
+  def handle_info({:assign_memory_usage, pid, node_id}, socket)
+      when not is_nil(pid) and not is_nil(node_id) do
+    Process.send_after(self(), {:assign_memory_usage, pid, node_id}, @memory_poll_interval_ms)
+
+    raw =
+      case :erlang.process_info(pid, :memory) do
```
kraleppa, line 132: Could you please create wrappers for these erlang function in `services/system` - we may want to mock them in the tests and there are some rumors that LiveView might get its own Debug module, so we want to have an easy way to swap them

kraleppa (commented):
Thanks, for the contribution - my only concern with this is the fact that I'm not sure if this information should be inside the sidebar from the UX perspective 🤔 Adding this info lefts less space for components tree and nested LiveViews. On the other hand I can't come up with a better place to display this information. 

I'll talk with our designer next week, if you had any other ideas about positioning them feel free to share them. Anyway this will have to wait till we deploy v0.2.0 (it may cause some conflicts)

lib/live_debugger/live_views/sidebar_live.ex
nathanmalishev: @kraleppa added in the wrappers. 

I did opt to skip the following in the ErlangService, but it can be easily added if you prefer.
```elixir
words = :erts_debug.size(assigns)
words * :erlang.system_info(:wordsize)
```

lib/live_debugger/services/system/erlang_service.ex
```diff
@@ -0,0 +1,49 @@
...
+      _ -> :error
+    end
+  end
+
+  @impl true
+  def process_info(_pid, _type), do: :error
+
+  @impl true
+  def garbage_collect(pid) do
+    :erlang.garbage_collect(pid)
```
kraleppa, line 47: I'm not sure if triggering garbage collection manually on the debugged application PID is a good idea. I have a feeling like this is a bit too big interference in the debugged app - at this point I have a feeling that LiveDebugger should purely "watch" how application behaves, not trigger any actions

PR #315, opened 2025-04-25
Refactor: Make LiveDebugger.Utils.PubSub mockable

hhubert6:


lib/live_debugger/services/system/module_service.ex
```diff
@@ -10,7 +10,7 @@ defmodule LiveDebugger.Services.System.ModuleService do
   @doc """
   Wrapper for `:code.all_loaded/0` that returns a list of loaded modules.
   """
-  @spec all() :: [{module(), charlist()}]
```
hhubert6, line 13: I found a mismatch here so I fixed it

GuzekAlan (approved):

PR #318, opened 2025-04-25
Task: Add mockable wrapper for :dbg

hhubert6:


lib/live_debugger/services/system/dbg_service.ex
```diff
@@ -0,0 +1,76 @@
+defmodule LiveDebugger.Services.System.DbgService do
+  @moduledoc """
+  This module provides wrappers for system functions that are used for debugging.
+  """
+
+  @callback tracer(:process, handlerSpec) :: {:ok, pid()} | {:error, error :: term()}
+            when handlerSpec: {handlerFun, initialData :: term()},
+                 handlerFun: (event :: term(), data :: term() -> newData :: term())
```
hhubert6, line 8: I don't know how to exactly write those specs. Is copying all of them from :dbg.tracer alright or just keep the ones we're actually using?

GuzekAlan: You can add types on top of module to reuse them and make them snake case.

```elixir
@type handler_function() :: (event :: term(), data :: term() -> term())

@callback tracer(:process, {handler_function(), initial_data :: term()) :: {:ok, pid()} | {:error, term()}
```

IMO good typing is nice so that we would know exactly what it returns (since it is a wrapper for build-in functions).

GuzekAlan (approved):

PR #328, opened 2025-04-28
Docs: Better design of hexdocs

kraleppa:


docs/config.md
```diff
@@ -0,0 +1,39 @@
...
+```elixir
+# config/dev.exs
+
+config :live_debugger,
+  ip: {127, 0, 0, 1}, # IP on which LiveDebugger will be hosted
+  port: 4007, # Port on which LiveDebugger will be hosted
+  secret_key_base: "YOUR_SECRET_KEY_BASE", # Secret key used for LiveDebugger.Endpoint
+  signing_salt: "your_signing_salt", # Signing salt used for LiveDebugger.Endpoint
+  adapter: Bandit.PhoenixAdapter # Adapter used in LiveDebugger.Endpoint
+  tracing_setup_delay: 0 # Time in ms after tracing will be initialized. Useful in case multi-nodes envs
```
GuzekAlan, line 38: missing `,` here (probably in README as well)

GuzekAlan (approved):

PR #337, opened 2025-04-29
Enhancement: added `server` option to config

kraleppa:


lib/live_debugger.ex
```diff
@@ -61,6 +61,15 @@ defmodule LiveDebugger do
         live_reload: Keyword.get(config, :live_reload, [])
       ]
 
+    endpoint_server = Keyword.get(config, :server)
```
GuzekAlan, line 64: Why this different than the rest of endpoint config (why didnt you use it with default falase and above insdead of using if statement)?

kraleppa: When you set it to `true` by default => LiveDebugger will always be run even when you start mix task like `mix ecto.migrate`

When you set it to `false` by default => LiveDebugger won't be started even when application is started with `iex -S mix phx.server`

Phoenix checks whether `server` config is present and if not it checks if `phx.server` flag was passed

Because of that we should only add this option when it is explicitly set to sth

GuzekAlan (approved):

PR #343, opened 2025-04-29
Task: Update issues templates

GuzekAlan:


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -1,7 +1,7 @@
 name: Bug Report
 description: Create a report to help us improve
+title: "Bug: "
```
kraleppa, line 3: Why do we add this prefix? We don't do it for features, and we have label for bugs

GuzekAlan: I thought that it would be good do make issues created by users distinct but since we have labels and triage it might be useless

kraleppa (approved):

PR #361, opened 2025-05-06
Enhancement: Create form for filtering by execution time

hhubert6:
Form design is not final. It's just for this feature to work.


lib/live_debugger_web/components.ex
```diff
@@ -72,6 +72,57 @@ defmodule LiveDebuggerWeb.Components do
...
+  Renders an input with label.
+  """
+  attr(:id, :any, default: nil)
+  attr(:name, :any)
+  attr(:label, :string, default: nil)
+  attr(:value, :any)
+
+  attr(:type, :string,
+    default: "text",
+    values: ~w(checkbox color date datetime-local email file month number password
```
kraleppa, line 85: Checkbox is already defined in the component above, also I'm not sure if this input will for each of these type. I'd consider removing allowed values from this field tbh

lib/live_debugger_web/components.ex
```diff
@@ -72,6 +72,57 @@ defmodule LiveDebuggerWeb.Components do
...
+               range search select tel text textarea time url week)
+  )
+
+  attr(:field, Phoenix.HTML.FormField, required: true)
+
+  attr(:rest, :global,
+    include: ~w(accept autocomplete capture cols disabled form list max maxlength min minlength
```
kraleppa, line 92: If Phoenix does not force you to declare these values here I would probably remove them

lib/live_debugger_web/components.ex
```diff
@@ -72,6 +72,57 @@ defmodule LiveDebuggerWeb.Components do
...
+  )
+
+  slot(:inner_block)
+
+  def input(%{field: %Phoenix.HTML.FormField{} = field} = assigns) do
+    assigns =
+      assigns
+      |> assign(field: nil, id: assigns.id || field.id)
+      |> assign_new(:name, fn -> field.name end)
+      |> assign_new(:value, fn -> field.value end)
```
kraleppa, line 103: You usually want this field to accept only `Phoenix.HTML.FormField` as attribute that stores all arguments passed to input

Check input definition from here
https://hexdocs.pm/phoenix_live_view/form-bindings.html#form-events

lib/live_debugger_web/components.ex
```diff
@@ -72,6 +72,57 @@ defmodule LiveDebuggerWeb.Components do
...
+      </label>
+      <input
+        type={@type}
+        name={@name}
+        id={@id}
+        value={Phoenix.HTML.Form.normalize_value(@type, @value)}
+        class={[
+          "mt-2 block w-full rounded-lg bg-surface-1-bg  focus:ring-0 text-xs",
+          "phx-no-feedback:border-zinc-300 phx-no-feedback:focus:border-zinc-400",
+          "border-default-border focus:border-secondary-text"
```
kraleppa, line 118: You should probably add `class` attribute to make sure that it is possible to change styling of this input as needed.

lib/live_debugger_web/live_components/filters_form.ex
```diff
@@ -21,7 +21,8 @@ defmodule LiveDebuggerWeb.LiveComponents.FiltersForm do
 
   @impl true
   def render(assigns) do
-    assigns = assign(assigns, :selected_filters_number, calculate_selected_filters(assigns.form))
+    assigns =
+      assign(assigns, :selected_filters_number, calculate_selected_filters(assigns.form) - 2)
```
kraleppa, line 25: Not sure how this calculation should work after these changes - we'll have to wait for the designs. However, for this moment I'd assume that it should calculate only marked checkboxes. To achieve that I'd change `calculate_selected_filters` function - it should probably filter out fields that are not callback checkboxes

lib/live_debugger_web/live_components/filters_form.ex
```diff
@@ -34,6 +35,11 @@ defmodule LiveDebuggerWeb.LiveComponents.FiltersForm do
                 <.checkbox field={@form[function]} label={"#{function}/#{arity}"} />
               <% end %>
             </div>
+            <p class="font-medium mb-4 mt-6">Callback execution time</p>
+            <div class="flex flex-col gap-3">
+              <.input label="max [us]" field={@form[:exec_time_max]} type="number" min="0" />
```
kraleppa, line 40: Check this `&micro;` 🇺🇸 🇺🇸 🇺🇸

kraleppa (commented):
I assume that this is form only that is not filtering yet, right? I'm asking since execution time does not appear in my case at all

https://github.com/user-attachments/assets/be2a3248-6ac4-4b15-8d2f-cc7c431a19d6

lib/live_debugger_web/live_components/filters_form.ex
GuzekAlan, line 1: How about adding a struct here? Also making min and max fields separate ? It can be done in another PR

GuzekAlan (approved):
image

You can pass greater min value than max value but since UI might change it can be resolved in later PR with connecting to actual filtering

lib/live_debugger_web/components.ex
```diff
@@ -72,6 +72,43 @@ defmodule LiveDebuggerWeb.Components do
...
+  attr(:input_class, :any, default: nil)
+  attr(:label_class, :any, default: nil)
+  attr(:rest, :global, include: ~w(min max))
+
+  slot(:label)
+
+  def input(assigns) do
+    ~H"""
+    <div phx-feedback-for={@field.name} class={["" | List.wrap(@wrapper_class)]}>
+      <label for={@field.id} class={["block font-medium text-xs" | List.wrap(@label_class)]}>
```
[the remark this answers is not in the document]
hhubert6: I tried to handle somehow that we want to display `&micro;`. When we pass it as string attribute then it needs to be rendered with `raw(...)`. Should I always use `raw` or use a flag (e.g. `label_raw`), for example like that
`<.input label_text="min [&micro;s]" label_raw />`

PR #364, opened 2025-05-07
Task: Implement caching mechanism

GuzekAlan:


lib/live_debugger/gen_servers/callback_tracing_server.ex
```diff
@@ -179,10 +179,13 @@ defmodule LiveDebugger.GenServers.CallbackTracingServer do
   defp do_publish(%{module: Phoenix.LiveView.Diff} = trace) do
     trace
     |> PubSubUtils.component_deleted_topic()
-    |> PubSubUtils.broadcast({:new_trace, trace})
+    |> PubSubUtils.broadcast({:component_deleted, trace})
```
kraleppa, line 182: Why don't you simply use `component_deleted` topic? Why do you broadcast this message to 2 channels?

lib/live_debugger/gen_servers/state_server.ex
```diff
@@ -0,0 +1,114 @@
...
+  def start_link(args \\ []) do
+    GenServer.start_link(__MODULE__, args, name: __MODULE__)
+  end
+
+  @impl true
+  def init(_args) do
+    :ets.new(@ets_table_name, [:named_table, :public, :ordered_set])
+
+    PubSubUtils.node_rendered()
+    |> PubSubUtils.subscribe!()
```
kraleppa, line 46: Imo you should subscribe here to `topic/socket_id/*/render`

lib/live_debugger/gen_servers/state_server.ex
```diff
@@ -0,0 +1,114 @@
...
+      publish_state_changed(trace, channel_state)
+    end
+  end
+
+  defp publish_state_changed(%Trace{} = trace, channel_state) do
+    socket_id = trace.socket_id
+    transport_pid = trace.transport_pid
+    node_id = trace.cid || trace.pid
+
+    PubSubUtils.state_changed_topic(socket_id, transport_pid, node_id)
```
kraleppa, line 90: I like the idea of creating more highlevel topic to detect whether change has been changed. Is there a chance to create a publish somehwere a diagram of routing? It gets complicated

lib/live_debugger/structs/lv_process.ex
```diff
@@ -59,7 +59,7 @@ defmodule LiveDebugger.Structs.LvProcess do
   """
   @spec new(pid :: pid()) :: t() | nil
   def new(pid) do
-    case LiveDebugger.Services.ChannelService.state(pid) do
+    case LiveDebugger.GenServers.StateServer.get(pid) do
```
kraleppa, line 62: You kept the wrapper for this function - `ChannelService.state/1`. Why aren't you using it?

lib/live_debugger_web/helpers/state_helper.ex
```diff
@@ -0,0 +1,20 @@
...
+  """
+  alias LiveDebugger.CommonTypes
+  alias LiveDebugger.Services.ChannelService
+
+  @doc """
+  Fetches state using ChannelService when no state is passed
+  """
+  @spec maybe_get_state(pid :: pid(), channel_state :: CommonTypes.channel_state() | nil) ::
+          {:ok, CommonTypes.channel_state()} | {:error, term()}
+  def maybe_get_state(pid, channel_state \\ nil) when is_pid(pid) do
```
kraleppa, line 13: What's the point of this function? Why should it be public?

GuzekAlan: To use it in 2 different modules. (same logic is used in different places)

lib/live_debugger/structs/lv_process.ex
GuzekAlan: I don't want to make Struct dependant on Service since it might lead to circular dependency but since it is alias I guess it might stay as it was. I thought there was `Process` function call but maybe something changed when I rebased

lib/live_debugger/gen_servers/callback_tracing_server.ex
GuzekAlan: We want to catch all `component_deleted` topics in `StateServer` since we want to update it for all applications which are alive (not only currently debugged one)

lib/live_debugger/gen_servers/state_server.ex
GuzekAlan: Yeah it might be a  good idea since it gave me a headache to properly connect them

lib/live_debugger/gen_servers/state_server.ex
GuzekAlan: I thought of making `___f_topic` but it looked really weird to me

kraleppa (approved):
Good job! I'd really like to see the diagram of routing

PR #373, opened 2025-05-09
Bug: fixed callback tracing after components switching

kraleppa:
Basically when you have callback tracing started and you change debugged node, callbacks from previous component will still appear in the callback traces

Before

https://github.com/user-attachments/assets/a08e1129-393a-4d67-aa96-2ccfe67cc88b


kraleppa:
After

https://github.com/user-attachments/assets/afe03c8d-9b8e-461f-9bc6-5c5a28a8e6a0

hhubert6 (approved):

GuzekAlan (approved):
Could you add check for this in TracingHelper or a comment in moduledoc about which assigns should be added for it to properly work ?

PR #382, opened 2025-05-13
Feature: add view with active LiveViews per window

kraleppa:
Also, when user using extension click return button they'll be redirected to that view


lib/live_debugger_web/hooks/iframe_detector.ex
```diff
@@ -0,0 +1,36 @@
...
+    |> attach_hook(:detect_iframe, :handle_event, &maybe_receive_detect_iframe/3)
+    |> cont()
+  end
+
+  defp maybe_receive_detect_iframe("detect-iframe", %{"in_iframe?" => in_iframe?}, socket) do
+    {:halt, assign(socket, :in_iframe?, in_iframe?)}
+  end
+
+  defp maybe_receive_detect_iframe(_, _, socket), do: {:cont, socket}
+end
```
GuzekAlan, line 36: This is nice handler to check this but doesn't it  leave a moment when we mount and render and only later correct if this is in iframe or not? Wouldn't it be better to expose different route (e.g. `/iframe/:pid` or adding `live_action` `:iframe` ? We thought about adding `AllowIframe` plug only for iframe so it  might combine this. What do you think?

kraleppa: It does leave a moment, but it's really small. I haven't found an easy way to check if request comes from iframe via plug. I tried adding a query param when request comes from extension and it worked until I haven't changed LiveView since we remove query params in such case. We'd have to change the routing system and rewrite this param every time. 

IMO it's simpler and this moment when url is incorrect is small enough to ignore it

kraleppa: Fixed it 👌

lib/live_debugger_web/live/channel_dashboard_live.ex
```diff
@@ -17,6 +16,14 @@ defmodule LiveDebuggerWeb.ChannelDashboardLive do
   alias LiveDebuggerWeb.SidebarLive
   alias LiveDebugger.Utils.PubSub, as: PubSubUtils
   alias LiveDebuggerWeb.Helpers.RoutesHelper
+  alias LiveDebuggerWeb.Assigns
+
+  @impl true
+  def mount(_params, _session, socket) do
+    socket
+    |> Assigns.assign_in_iframe?()
```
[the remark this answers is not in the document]
kraleppa: Yup, we can 👌

lib/live_debugger_web/components/tab_group.ex
```diff
@@ -0,0 +1,88 @@
...
+
+  attr(:transport_pid, :any, required: true)
+  attr(:grouped_lv_processes, :list, required: true)
+  attr(:show_transport_pid?, :boolean, default: true)
+
+  def group(assigns) do
+    ~H"""
+    <div class="w-full h-max flex flex-col shadow-custom rounded-sm bg-surface-2-bg border border-default-border">
+      <div
+        :if={@show_transport_pid?}
```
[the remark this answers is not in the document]
kraleppa: No, not really 👍

PR #383, opened 2025-05-13
POC: Inspecting LiveVIew

GuzekAlan:


assets/js/client.js
```diff
@@ -46,6 +47,10 @@ function highlightingEnabled() {
   }
 }
 
+function inspectEnabled() {
+  return true;
```
kraleppa, line 51: What this function is supposed to do in the future? Is it a feature flag?

kraleppa (commented):
I really like your approac - it's really creative and having components inspection working not only for DevTools is game-changer. 

However I have thought. We are already sending events via debugged application websocket (associated with highlighting). Now we'll have even more events coming from server to client and we also have client server communication (the post request you make)

Maybe we should create a separate websocket for LiveDebugger communication and move whole LiveDebugger communication there? In the browser we should have phoenix socket already defined and we should be able to easily declare a new socket on the server side (since we have a new endpoint)

mix.exs
```diff
@@ -67,6 +67,7 @@ defmodule LiveDebugger.MixProject do
       {:phoenix_live_view, "~> 0.20 or ~> 1.0"},
       {:igniter, "~> 0.5 and >= 0.5.40", optional: true},
       {:bandit, "~> 1.6", only: [:dev, :test]},
+      {:cors_plug, "~> 3.0"},
```
kraleppa, line 70: Is it really necessary to add this dep?

assets/js/client.js
GuzekAlan: This is draft project to show how it is possible to make it work. I didn't made the final changes here. I will create proper PR in task with implementation

kraleppa: ok, just wanted to understand what is happening here 👍

GuzekAlan:
@kraleppa 

> Maybe we should create a separate websocket for LiveDebugger communication and move whole LiveDebugger communication there? In the browser we should have phoenix socket already defined and we should be able to easily declare a new socket on the server side (since we have a new endpoint)

As I understand we will create a separate socket which will be registered using our `client.js` which is conditional based on `browser_features` config. It is a good idea. We could create `Phoenix.Channel` even for this type of streaming events which might make highlighting and inspecting more smooth.

PR #388, opened 2025-05-14
Task: Adjust devtools extension for firefox

hhubert6:


GuzekAlan (approved):

devtools/firefox/devtools.js
```diff
@@ -0,0 +1,81 @@
+function getLiveDebuggerSessionURL() {
+  return new Promise((resolve, reject) => {
+    const script = `
```
kraleppa, line 3: Maybe this script could be shared between those two extensions?

kraleppa (approved):

devtools/common/url.js
```diff
@@ -0,0 +1,51 @@
...
+        return getSessionURL(baseURL);
+      })();
+    `;
+
+    browser.devtools.inspectedWindow.eval(script, (result, isException) => {
+      if (isException || !result) {
+        reject(new Error("Error fetching LiveDebugger session URL"));
+      } else {
+        resolve(result);
+      }
```
GuzekAlan, line 48: `browser` should be in common ?

hhubert6: It's a function argument

GuzekAlan: Then could you please name it differently. This is JS and I am worried about what might happen

GuzekAlan (approved):

PR #412, opened 2025-05-20
Task: Add mode for disconnected LiveViews

GuzekAlan:


lib/live_debugger_web/endpoint.ex
```diff
@@ -18,22 +18,31 @@ defmodule LiveDebuggerWeb.Endpoint do
...
+        gzip: false
+      )
+
+    true ->
+      plug(Plug.Static,
+        at: "/assets/live_debugger",
+        from: {:live_debugger, "priv/static"},
+        gzip: false
+      )
   end
```
GuzekAlan, line 46: if we will use `priv/static` for tests it might not pass on development (that was my case). But if we will run tests on `priv/static/dev` then we are not testing "production" environment. But since both these files are the same i think we shouldn't worry about it

README.md
```diff
@@ -97,6 +97,18 @@ In `router.ex` of your Phoenix app, make sure your locally running Phoenix app c
     plug :put_secure_browser_headers, %{"content-security-policy" => @csp}
 ```
 
+### Dead View Mode
```
kraleppa, line 100: 1. It cannot be merged to README until we release v0.3.0
2. Currently you describe a feature in `Getting Started Section`
3. Tbh I'd move it to this section above the movie where features are listed in a short way
4. If we want to describe dead view mode in a more detailed way I think we do it inside hexdocs
5. I think we can describe each feature (by every I mean the debugging features) in hexdocs

lib/live_debugger/env.ex
```diff
@@ -19,4 +19,8 @@ defmodule LiveDebugger.Env do
     Application.get_env(:live_debugger, :test_mode?, false) and
       not Application.get_env(:live_debugger, :e2e?, false)
   end
+
+  def dead_view_mode?() do
```
kraleppa, line 23: We have `lib/live_debugger/feature.ex` for this kind of stuff

lib/live_debugger/gen_servers/ets_table_server.ex
```diff
@@ -1,33 +1,54 @@
...
+
+  It sends `{:process_status, {:dead, pid}}` to the process status topic.
+
+  ## Dead View Mode
+  When in dead view mode, the gen_server will send `{:process_status, {:dead, pid}}` to the process status topic when a process dies.
+  It will wait for all watchers to be removed and then delete the ETS table and sends `{:process_status, {:dead, pid}}` to the process status topic.
   """
 
+  defmodule TableInfo do
+    @moduledoc false
```
kraleppa, line 13: You can describe here each field

lib/live_debugger_web/components/navbar.ex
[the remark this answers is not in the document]
kraleppa: I totally agree - very good refactor btw

lib/live_debugger_web/components/navbar.ex
```diff
@@ -0,0 +1,147 @@
...
+  @doc """
+  Component for displaying the connection status of a LiveView.
+  When button is clicked, it will trigger a `find-successor` event with the PID of the LiveView.
+  """
+  attr(:id, :string, required: true)
+  attr(:connected?, :boolean, required: true, doc: "Whether LiveView is connected.")
+  attr(:pid, :string, required: true, doc: "The PID of the LiveView.")
+  attr(:rest, :global)
+
+  def connected(assigns) do
```
kraleppa, line 117: We need some kind of tooltip here tbh

kraleppa (commented):
Also, we'll probably need a checkbox in settings to disable this mode. But this is for the future, after I implement the sidebar

devtools/common/devtools.html
```diff
@@ -1,11 +1,14 @@
...
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+</head>
+
+<body>
+  <script src="url.js"></script>
+  <script src="devtools.js"></script>
+</body>
+
+</html>
```
kraleppa, line 14:
```suggestion
</html>

```

lib/live_debugger.ex
```diff
@@ -86,6 +86,7 @@ defmodule LiveDebugger do
     debug_button? = Keyword.get(config, :debug_button?, true)
     highlighting? = Keyword.get(config, :highlighting?, true)
     version = Application.spec(:live_debugger)[:vsn] |> to_string()
+    devtools_allow_redirects = Keyword.get(config, :devtools_allow_redirects, true)
```
kraleppa, line 89: I'd go with false here tbh

kraleppa (approved):

PR #422, opened 2025-05-22
Task: Update callback execution time info according to designs

hhubert6:


kraleppa (approved):
> How about parsing the timer?

That's fine, I just don't want it to shake 😄

PR #432, opened 2025-05-26
Bug: All traces are loaded when no callback name is checked in filters

hhubert6:


kraleppa:
@hhubert6 after merge you'll need to cherry-pick it to `v0.2` branch

PR #433, opened 2025-05-26
Enhancement: apply new navigation to UI

kraleppa:


GuzekAlan (commented):
I don't know if this is in this PR but there shouldn't be icon to return on iframe view

image


There is problem with long names in sidebar. Some flexbox/truncate tinkering must be done

image


I assume that tooltips for navigation left bar will be added later

GuzekAlan (approved):
Nice views :)

PR #435, opened 2025-05-27
Feature: add settings panel 

kraleppa:


test/live_debugger/channel_dashboard_test.exs
```diff
@@ -108,6 +108,21 @@ defmodule LiveDebugger.ChannelDashboardTest do
...
+    dev_app
+    |> visit(@dev_app_url)
+
+    debugger
+    |> visit("/")
+    |> click(first_link())
+    |> assert_has(css("navbar a[href=\"/settings\"]"))
+    |> click(css("navbar a[href=\"/settings\"]"))
+    |> assert_has(css("h1", text: "Settings"))
+  end
```
GuzekAlan, line 124: we could also add case when returning from settings returns to appropriate page

lib/live_debugger_web/live/settings_live.ex
```diff
@@ -0,0 +1,162 @@
...
+                </div>
+              </div>
+            </div>
+          </div>
+        </div>
+        <div class="mt-6">
+          <div class="bg-surface-0-bg rounded shadow-custom border border-default-border">
+            <div class="p-6">
+              <div class="flex flex-col md:flex-row justify-between md:items-center gap-4">
+                <div class="flex flex-col gap-1">
```
GuzekAlan, line 83: is it necessary to have such nested divs?

test/live_debugger/channel_dashboard_test.exs
kraleppa: Tbh I'd wait for it since we have discussion on figma how exactly this return arrow should work - we'll fix it in the separate PR

PR #438, opened 2025-05-27
Task: Add custom LiveDebugger url config

GuzekAlan:

PR #442, opened 2025-05-28
Backport: Add custom LiveDebugger url config

GuzekAlan:
* hide behind config option

* change name, update docs

* fix tests

### DIFF
--- a/lib/live_debugger.ex
+++ b/lib/live_debugger.ex
@@ -80,13 +80,14 @@ defmodule LiveDebugger do
   defp put_live_debugger_tags(config) do
     ip_string = config |> Keyword.get(:ip, @default_ip) |> :inet.ntoa() |> List.to_string()
     port = Keyword.get(config, :port, @default_port)
+    live_debugger_url = "http://#{ip_string}:#{port}"
 
     browser_features? = Keyword.get(config, :browser_features?, true)
     debug_button? = Keyword.get(config, :debug_button?, true)
     highlighting? = Keyword.get(config, :highlighting?, true)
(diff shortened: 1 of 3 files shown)


lib/live_debugger.ex
```diff
@@ -80,13 +80,14 @@ defmodule LiveDebugger do
...
+    live_debugger_url = "http://#{ip_string}:#{port}"
 
     browser_features? = Keyword.get(config, :browser_features?, true)
     debug_button? = Keyword.get(config, :debug_button?, true)
     highlighting? = Keyword.get(config, :highlighting?, true)
+    external_url = Keyword.get(config, :external_url, live_debugger_url)
 
-    live_debugger_url = "http://#{ip_string}:#{port}"
-    live_debugger_assets_url = "http://#{ip_string}:#{port}/#{@assets_path}"
+    live_debugger_assets_url = "#{external_url}/#{@assets_path}"
```
rhcarvalho, line 90: Trying to repeat my review from https://github.com/software-mansion/live-debugger/pull/438#pullrequestreview-2874288801

Looks like GitHub posted only the overall comment and lost the actual review comment.

cc @kraleppa 

What I was trying to point out is that I think the correct solution would be to set `live_debugger_assets_url` based on `live_debugger_url` as before, and allow `live_debugger_url` to be overridable -- either as a string like now or in a more granular fashion like `Phoenix.Endpoint` config.

The problem with not setting the `:url` assign according to what the browser sees is that it gets used in `SidebarLive` for a `push_patch`.

kraleppa (approved):

PR #447, opened 2025-05-29
Enhancement: global traces preparations

kraleppa:
Since this PR is already quite big and there is much more left to do I decided to split it into parts for easier review:
- I made some preparations for routing (we'll need channel that does not include `node_id`)
- Extended tooltips
- Created template for global traces with with working routing

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -40,17 +40,9 @@ body::-webkit-scrollbar {
 }
 
 .tooltip-primary {
-  background-color: var(--button-primary-bg);
-  color: var(--button-primary-content);
-  padding: 0.5rem;
-  border-radius: 0.25rem;
   font-weight: 600;
 }
 
 .tooltip-secondary {
-  background-color: var(--button-secondary-bg);
-  color: var(--button-secondary-content);
-  padding: 0.5rem;
-  border-radius: 0.25rem;
-  font-weight: 600;
+  font-weight: 400;
 }
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -38,3 +38,19 @@ body::-webkit-scrollbar {
 ::-webkit-scrollbar-corner {
   background: transparent;
 }
+
+.tooltip-primary {
+  background-color: var(--button-primary-bg);
+  color: var(--button-primary-content);
+  padding: 0.5rem;
+  border-radius: 0.25rem;
+  font-weight: 600;
+}
+
+.tooltip-secondary {
+  background-color: var(--button-secondary-bg);
+  color: var(--button-secondary-content);
+  padding: 0.5rem;
+  border-radius: 0.25rem;
+  font-weight: 600;
+}
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -38,11 +38,3 @@ body::-webkit-scrollbar {
 ::-webkit-scrollbar-corner {
   background: transparent;
 }
-
-.tooltip-primary {
-  font-weight: 600;
-}
-
-.tooltip-secondary {
-  font-weight: 400;
-}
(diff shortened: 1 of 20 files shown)


lib/live_debugger_web/live/global_traces_live.ex
```diff
@@ -0,0 +1,71 @@
...
+    """
+  end
+
+  defp get_return_link(lv_process, in_iframe?) do
+    cond do
+      not in_iframe? ->
+        RoutesHelper.live_views_dashboard()
+
+      not lv_process.ok? ->
+        nil
```
GuzekAlan, line 65: why shouldn't user be able to return when they are loading `lv_process` ??

lib/live_debugger_web/live/global_traces_live.ex
```diff
@@ -0,0 +1,71 @@
...
+          return_link={get_return_link(@lv_process, @in_iframe?)}
+          current_url={@url}
+          class="sm:hidden"
+        />
+        <Navbar.live_debugger_logo_icon />
+        <div
+          :if={not @lv_process.ok?}
+          class="animate-pulse w-36 bg-surface-1-bg rounded text-surface-1-bg"
+        >
+          Loading...
```
GuzekAlan, line 29: Why text here if it is invisible ?

lib/live_debugger_web/helpers/routes_helper.ex
```diff
@@ -12,9 +12,26 @@ defmodule LiveDebuggerWeb.Helpers.RoutesHelper do
...
-    ~p"/pid/#{Parsers.pid_to_string(pid)}"
+    pid
+    |> Parsers.pid_to_string()
+    |> channel_dashboard()
+  end
+
+  def channel_dashboard(pid) when is_binary(pid) do
+    ~p"/pid/#{pid}"
+  end
+
```
GuzekAlan, line 25: we might rename it from `ChannelDashboard` to NodeInspector` (not necessarily in this PR)

lib/live_debugger_web/components/navigation_menu.ex
```diff
@@ -39,10 +72,34 @@ defmodule LiveDebuggerWeb.Components.NavigationMenu do
...
+    url
+    |> String.split("/")
+    |> Enum.at(3, "node_inspector")
+  end
+
+  defp get_pid(url) do
+    url
+    |> String.split("/")
+    |> Enum.at(2)
+  end
```
GuzekAlan, line 104: since url might have query params please add functionality in `LiveDebugger.Utils.URL`. As I tested it crushes when changing to global traces with LiveComponent being selected

assets/css/app.css
```diff
@@ -38,3 +38,19 @@ body::-webkit-scrollbar {
 ::-webkit-scrollbar-corner {
   background: transparent;
 }
+
+.tooltip-primary {
+  background-color: var(--button-primary-bg);
+  color: var(--button-primary-content);
+  padding: 0.5rem;
+  border-radius: 0.25rem;
+  font-weight: 600;
```
GuzekAlan, line 47: since these are the same in both cases you could add it to html element with tooltip

kraleppa: Why would we style something `inline` by modifying element via JS if we have css classes that can contain these styles?

assets/css/app.css
```diff
@@ -38,3 +38,19 @@ body::-webkit-scrollbar {
...
+}
+
+.tooltip-secondary {
+  background-color: var(--button-secondary-bg);
+  color: var(--button-secondary-content);
+  padding: 0.5rem;
+  border-radius: 0.25rem;
+  font-weight: 600;
+}
```
hhubert6, line 56: It's been discussed with the designer that all tooltips should have the same colors, see figma.

lib/live_debugger_web/components/navigation_menu.ex
GuzekAlan: I mean to have functionality which takes `nth` element from url's path and use it here (to not have problems with query params etc)

PR #465, opened 2025-06-03
Bug: Flaky tests in state server

GuzekAlan:
Because there are multiple tests which operate on same name ets table they should be sync. Another approach would be mocking function which returns ets table name but then we would need to expose it (as non-private) and since these tests are small we might not need to do it. 
If you think that it should be mocked I'll change it

@kraleppa @hhubert6


GuzekAlan:
https://github.com/software-mansion/live-debugger/actions/runs/15413570603/job/43370980057

This couldn't be reproduced again. The proposed solution didn't fix actual problem

PR #488, opened 2025-06-06
Enhancement: applied new designs to filters form

kraleppa:


lib/live_debugger_web/components.ex
```diff
@@ -396,7 +396,7 @@ defmodule LiveDebuggerWeb.Components do
       id={@id}
       phx-hook="Fullscreen"
       class={[
-        "relative h-max w-full xl:w-max xl:min-w-[50rem] bg-surface-0-bg p-2 overflow-auto hidden flex-col rounded-md backdrop:bg-black backdrop:opacity-50"
+        "relative h-max w-full xl:w-max xl:min-w-[50rem] bg-surface-0-bg pt-1 overflow-auto hidden flex-col rounded-md backdrop:bg-black backdrop:opacity-50"
```
GuzekAlan, line 399: actually it is 4th time i see change in padding inside fullscreen. Maybe we should delete it and let parent components worry about styling it. What do you think?

lib/live_debugger_web/live/traces/components/clear_button.ex
```diff
@@ -31,9 +31,14 @@ defmodule LiveDebuggerWeb.Live.Traces.Components.ClearButton do
...
+    <.button
+      phx-click="clear-traces"
+      class="flex gap-2 [&>*:nth-child(2)]:ml-1"
+      variant="secondary"
+      size="sm"
+    >
       <.icon name="icon-trash" class="w-4 h-4" />
-      <div class={@label_class}>
+      <div class={[@label_class, "ml-1"]}>
         Clear
```
GuzekAlan, line 42: why both `nth-child` and actual child has `ml-1`?

lib/live_debugger_web/live_components/filters_form.ex
```diff
@@ -22,24 +22,34 @@ defmodule LiveDebuggerWeb.LiveComponents.FiltersForm do
...
+          </div>
+          <.filters_group_header
+            title="Execution Time"
+            class="pt-2"
+            group_name={:execution_time}
+            form={@form}
+            default_filters={@default_filters}
+            target={@myself}
+          />
+          <div class="pb-5">
```
GuzekAlan, line 51: Can't this div be flattened ?

lib/live_debugger_web/live_components/filters_form.ex
```diff
@@ -128,6 +140,21 @@ defmodule LiveDebuggerWeb.LiveComponents.FiltersForm do
     assign(socket, :form, form)
   end
 
+  def assign_form(socket, filters_list, form) when is_list(filters_list) do
```
GuzekAlan, line 143: You have form in socket.
Also shouldn't it be private?

lib/live_debugger_web/live/traces/components/clear_button.ex
kraleppa: oops `nth` child shouldn't be here

lib/live_debugger_web/live_components/filters_form.ex
kraleppa: This diff reperesent a section for time inputs and its errors

GuzekAlan (commented):
I've noticed that chaning filters and exiting fullscreen saves them (but don't update). It is misleading UX. We should always show current filters when opeing filters in fullscreen

PR #491, opened 2025-06-10
Feature: filters sidebar in global callback traces view

kraleppa:


GuzekAlan (commented):
When shrinking end expanding width of global traces view there are different form fields. But it looks like problem with sidebar which we should resolve in #495 

image

image


From UI/UX side I miss the button for not applying filters on window view. I've commented it on Figma

GuzekAlan:
What I meant here
> From UI/UX side I miss the button for not applying filters on window view. I've commented it on Figma

Was to add `Cancel` button to return state to currently selected filters

PR #494, opened 2025-06-10
Enhancement: Add module in label for global traces

GuzekAlan:


kraleppa (approved):
Could you please add a simple test to `test/live_debugger/process_callback_traces_test.exs` that checks the module name?

Issue #495, opened 2025-06-10
Refactor: Change Sidebar to not have duplicated children


GuzekAlan:
Currently we need to specify 2 elements for sidebar: one for window view and one for small mobile view. It leads to problems with forms, collapsibles.

The goal of this task is to figure out and implement the way to achieve having only one dom element for content of sidebar in our app.

samrobinsonsauce:
Is it a requirement to keep the sidebar open when transitioning to smaller screen sizes?

Could we not just hide it and have the user re-open it if needed?

PR #506, opened 2025-06-12
consolidate sidebar children

samrobinsonsauce:
Potential fix for https://github.com/software-mansion/live-debugger/issues/495.

Renders a single `<.sidebar_content>` component rather than duplicates which helps retain form state when we transition to a mobile view.


hhubert6 (commented):
Looks good, but it definitely needs to be incorporated into `sidebar_slide_over` so that it could be reusable.

kraleppa (approved):
awesome - thanks for contribution!

GuzekAlan (approved):

hhubert6 (approved):

PR #515, opened 2025-06-24
Bug: fix scrollbar size on firefox

kraleppa:


assets/css/app.css
```diff
@@ -46,42 +46,17 @@
...
 }
 
-.remove-arrow::-webkit-inner-spin-button,
-.remove-arrow::-webkit-outer-spin-button {
-  -webkit-appearance: none;
-  margin: 0;
-}
-.remove-arrow {
-  -moz-appearance: textfield;
-}
```
hhubert6, line 68: This class is used to remove arrows in input type number which are not very good looking and they are certainly not present in designs. Did you remove this on purpose?

PR #565, opened 2025-07-07
Refactor: Create `LiveDebugger.API.System.Module`

hhubert6:


lib/live_debugger_refactor/api/system/module.ex
```diff
@@ -0,0 +1,56 @@
+defmodule LiveDebugger.API.System.Module do
```
kraleppa, line 1: I think we should stick with naming new modules `LiveDebuggerRefactor` to make sure that we won't have a situation when we duplicate module name

kraleppa (approved):

PR #568, opened 2025-07-07
Refactor: Create `LiveDebugger.API.System.Process`

hhubert6:


lib/live_debugger_refactor/api/system/process.ex
```diff
@@ -0,0 +1,66 @@
+defmodule LiveDebugger.API.System.Process do
```
kraleppa, line 1: Same as previous PR
```suggestion
defmodule LiveDebuggerRefactor.API.System.Process do
```

lib/live_debugger_refactor/api/system/process.ex
```diff
@@ -0,0 +1,66 @@
...
+  """
+  @callback initial_call(pid :: pid()) :: mfa() | nil
+  @callback state(pid :: pid()) :: {:ok, term()} | {:error, term()}
+  @callback list() :: [pid()]
+
+  @doc """
+  Wrapper for `Process.info/2` with some additional logic that returns the initial call of the process.
+  """
+  @spec initial_call(pid :: pid()) :: mfa() | nil
+  def initial_call(pid), do: impl().initial_call(pid)
```
kraleppa, line 13: Wondering if this function shouldn't return some kind of error e.g. `{:error, :not_alive} 🤔 IMO it describes a bit better an outcome of this function

test/live_debugger_refactor/api/system/process_test.exs
```diff
@@ -0,0 +1,48 @@
...
+    test "returns nil for a dead process", %{dead_pid: dead_pid} do
+      assert Process.alive?(dead_pid) == false
+      assert ProcessImpl.initial_call(dead_pid) == nil
+    end
+  end
+
+  describe "state/1" do
+    test "returns state of for a live process", %{alive_pid: alive_pid} do
+      assert {:ok, state} = ProcessImpl.state(alive_pid)
+      assert %{number: 14} = state
```
kraleppa, line 41: I think you should be able to do sth like this
```suggestion
      assert {:ok, %{number: 14}} = ProcessImpl.state(alive_pid)
```

PR #576, opened 2025-07-09
Refactor: Create `LiveDebuggerRefactor.API.TracesStorage`

hhubert6:
I will add tests for this module in a separate PR so this one wouldn't get too big. #580

### DIFF
--- a/lib/live_debugger_refactor/api/traces_storage.ex
+++ b/lib/live_debugger_refactor/api/traces_storage.ex
@@ -1,0 +1,127 @@
+defmodule LiveDebuggerRefactor.API.TracesStorage do
+  @moduledoc """
+  API for managing traces storage. In order to properly use invoke `init/0` at the start of application.
+  It uses Erlang's ETS (Erlang Term Storage).
+  """
+
+  alias LiveDebugger.Structs.Trace
+  alias LiveDebugger.CommonTypes
+
+  @type ets_continuation() :: term()
+  @typedoc """
+  Pid is used to store mapping to table references.
+  It identifies ETS tables managed by EtsTableServer.
+  """
+  @type ets_table_id() :: pid()
+  @type table_identifier() :: ets_table_id() | reference()
+
+  @callback init() :: :ok
+  @callback insert(Trace.t()) :: true
+  @callback insert(table_ref :: reference(), Trace.t()) :: true
+  @callback get(table_identifier(), trace_id :: integer()) :: Trace.t() | nil
+  @callback get(table_identifier(), opts :: keyword()) ::
+              {[Trace.t()], ets_continuation()} | :end_of_table
+  @callback clear(table_identifier(), node_id :: pid() | CommonTypes.cid() | nil) :: true
+  @callback clear(table_identifier()) :: true
+  @callback get_table(ets_table_id()) :: reference() | nil
+  @callback get_all_tables() :: [reference()]
+
+  defguard is_table_identifier(id) when is_pid(id) or is_reference(id)
+
+  @doc """
+  Initializes ets table.
+  It should be called when application starts.
+  """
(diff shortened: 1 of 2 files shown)


lib/live_debugger_refactor/api/traces_storage.ex
```diff
@@ -0,0 +1,391 @@
...
+  @callback insert(Trace.t()) :: true
+  @callback insert(table_ref :: reference(), Trace.t()) :: true
+  @callback get_by_id(table_identifier(), trace_id :: integer()) :: Trace.t() | nil
+  @callback get(table_identifier(), opts :: keyword()) ::
+              {[Trace.t()], ets_continuation()} | :end_of_table
+  @callback clear(table_identifier(), node_id :: pid() | CommonTypes.cid() | nil) :: true
+  @callback clear(table_identifier()) :: true
+  @callback get_table(ets_table_id()) :: reference()
+  @callback delete_table(table_identifier()) :: boolean()
+  @callback get_all_tables() :: [reference()]
```
kraleppa, line 28: I think some of these functions can raise errors (e.g. when you provide incorrect table ref. For such functions it is recommended to use `bang!` convention https://hexdocs.pm/elixir/main/naming-conventions.html#trailing-bang-foo

lib/live_debugger_refactor/api/traces_storage.ex
```diff
@@ -0,0 +1,391 @@
...
+
+  @doc """
+  Initializes ets table.
+  It should be called when application starts.
+  """
+  @spec init() :: :ok
+  def init(), do: impl().init()
+
+  @doc """
+  Inserts a new trace into the storage.
```
GuzekAlan, line 40: could you add comment on how they differ in performance (that using table_ref is faster but you need to have ref saved) ?

lib/live_debugger_refactor/api/traces_storage.ex
```diff
@@ -0,0 +1,391 @@
...
+    impl().clear(table_id, node_id)
+  end
+
+  @doc """
+  Deletes all traces for a given table.
+  """
+  @spec clear(table_identifier()) :: true
+  def clear(table_id) when is_table_identifier(table_id) do
+    impl().clear(table_id)
+  end
```
GuzekAlan, line 92: do we want to have deleting all traces implemented in 2 ways (by passing `nil` and by calling 1-arity function) ?

lib/live_debugger_refactor/api/traces_storage.ex
```diff
@@ -0,0 +1,391 @@
...
+
+        [{_, ref}] ->
+          :ets.delete(@processes_table_name, pid)
+          :ets.delete(ref)
+      end
+    end
+
+    @impl true
+    def get_all_tables() do
+      :ets.tab2list(@processes_table_name)
```
GuzekAlan, line 243: If we are exposing it as an API how about converting it to map? Or do you think list of tuples will be more versatile?

lib/live_debugger_refactor/api/traces_storage.ex
hhubert6: It is implemented that way in old api and it seems it's because of the way it is used in UI where there is passed node_id as assign which can be `nil` (in global traces view). Effectively `clear/1` is used only by `clear/2` but it might be used elsewhere depending on changes in UI due to refactor. I wasn't sure what to do about it so I left it as it is.

lib/live_debugger_refactor/api/traces_storage.ex
hhubert6: The only case in which it will be used that I thought of was by GarbageCollector which will be just iterating over it. That's why I left it as a list but I can change it if you want.

PR #578, opened 2025-07-10
Refactor: create base for each service

kraleppa:


lib/live_debugger.ex
```diff
@@ -25,6 +25,21 @@ defmodule LiveDebugger do
...
+    else
+      get_legacy_children(config)
+    end
+  end
+
+  defp get_refactor_children() do
+    []
+    |> LiveDebuggerWeb.Endpoint.append_endpoint_children()
+    |> LiveDebuggerRefactor.Bus.append_bus_tree()
+    |> LiveDebuggerRefactor.Services.append_services_children()
```
hhubert6, line 39: Maybe if out services so they don't start gen servers when unit testing? Like we did previously.
For example when testing Bus you start the pubsub server manually in tests what I think should be done for all other server when needed.

hhubert6 (commented):
Tests for `Bus` and `Event` have files names that do not end with `_test.exs`. Because of that they don't run when running `mix test` and on CI (I don't know if that was done intentionally?). Tests for `Bus` fail due to changes in this PR.

PR #579, opened 2025-07-10
Refactor: Create API for `StatesStorage`

GuzekAlan:


hhubert6 (approved):
If we want GarbageCollector to not use `:ets` functions directly we could add something like `delete/1`, `get_all_states` or `trim`. But it can be done later when implementing GC.

PR #588, opened 2025-07-14
Refactor: add tracing manager genserver

kraleppa:


lib/live_debugger_refactor/services/callback_tracer/process/tracer.ex
```diff
@@ -0,0 +1,11 @@
+defmodule LiveDebuggerRefactor.Services.CallbackTracer.Process.Tracer do
+  @moduledoc """
+  This module defines a function that used in the `:dbg.tracer` process.
```
hhubert6, line 3:
```suggestion
  This module defines a function that is used in the `:dbg.tracer` process.
```

test/live_debugger_refactor/bus_test.exs
```diff
@@ -128,4 +128,60 @@ defmodule LiveDebuggerRefactor.BusTest do
...
   end
+
+  describe "bang receive functions" do
+    test "receive_events! returns :ok on success" do
+      assert BusImpl.receive_events!() == :ok
+    end
+
+    test "receive_events! with pid returns :ok on success" do
+      assert BusImpl.receive_events!(self()) == :ok
+    end
```
hhubert6, line 139: Are those one line tests necessary if tests below already do the same?

```elixir
    test "bang functions work with broadcasting and receiving" do
      assert BusImpl.receive_events!() == :ok
      assert BusImpl.receive_events!(self()) == :ok

      assert BusImpl.broadcast_event!(%TestEvents.TestEvent{name: "test"}, self()) == :ok

      assert_receive %TestEvents.TestEvent{name: "test"}
      assert_receive %TestEvents.TestEvent{name: "test"}
    end
```

lib/live_debugger_refactor/services/callback_tracer/queries/callbacks.ex
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  defp debugger?(module) do
+    stringified_module = Atom.to_string(module)
+
+    String.starts_with?(stringified_module, [
+      "Elixir.LiveDebugger.",
+      "Elixir.LiveDebuggerWeb.",
+      "LiveDebugger.",
+      "LiveDebuggerWeb."
```
GuzekAlan, line 56: Maybe adding `LiveDebuggerRefactor` instead of it? This will be run with `LiveDebuggerRefactor` modules

lib/live_debugger_refactor/services/callback_tracer/gen_servers/tracing_manager.ex
```diff
@@ -0,0 +1,88 @@
...
+
+    {:noreply, state}
+  end
+
+  @impl true
+  def handle_info(%TracingRefreshed{}, state) do
+    apply_trace_patterns()
+
+    {:noreply, state}
+  end
```
GuzekAlan, line 70: If we have convention that event is sent after something occurs I would see that `TracingRefreshed` is send after `apply_trace_patterns`. It would add more logic so I guess it is okay but I don't know if this won't be misleading in the future

kraleppa: TracingRefreshed will be executed as result of action in UI called `refresh_tracing`. Here we don't have a proper action and if we had the event would be called `TracePatternsApplied`. We cannot avoid that some events will be similar to each other, but because event is defined in given module its name can help to which context given event belongs (e.g. `TracingRefreshed` full name would be probably something like `LiveDebugger.App.Settings.TracingRefreshed` which should give a pretty good knowledge where this event is sent)

PR #591, opened 2025-07-15
Refactor: Move general UI modules and lay foundation for UI

GuzekAlan:


lib/live_debugger_refactor/app/web.ex
```diff
@@ -0,0 +1,61 @@
...
+      import unquote(__MODULE__).Hooks.Flash, only: [push_flash: 2, push_flash: 3]
+    end
+  end
+
+  def component do
+    quote do
+      use Phoenix.Component
+
+      import Phoenix.HTML
+      import.Helpers
```
hhubert6, line 33:
```suggestion
      import unquote(__MODULE__).Helpers
```

lib/live_debugger_refactor/app/web/components.ex
```diff
@@ -0,0 +1,719 @@
+defmodule LiveDebuggerRefactor.App.Web.Components do
+  @moduledoc """
+  This module provides reusable components for LiveDebugger.
+  """
+
+  use Phoenix.Component
+
+  alias LiveDebugger.Utils.Parsers
```
hhubert6, line 8: Shouldn't you copy `Parsers` to `LiveDebuggerRefactor` as well?
```suggestion
  alias LiveDebuggerRefactor.Utils.Parsers
```

lib/live_debugger_refactor/app/web/layout.ex
```diff
@@ -0,0 +1,87 @@
...
+        <%= @inner_content %>
+      </body>
+    </html>
+    """
+  end
+
+  def render("app.html", assigns) do
+    ~H"""
+    <main class="h-screen w-screen max-w-full">
+      <LiveDebuggerWeb.Components.flash flash={@flash} />
```
hhubert6, line 66:
```suggestion
      <LiveDebuggerRefactor.App.Web.Components.flash flash={@flash} />
```

lib/live_debugger_refactor/app/web/helpers.ex
kraleppa, line 1: Wondering if this couldn't be moved to the module that Hubert provided https://github.com/software-mansion/live-debugger/blob/accf1b2d78685cd8b122a5d8e2e32d5b14174741/lib/live_debugger_refactor/services/helpers.ex

kraleppa (approved):

lib/live_debugger_refactor/app/web/helpers.ex
GuzekAlan: I thought about using specs to ensure that we return proper value from callbacks but it doesn't seem to have much advantage

hhubert6 (approved):

lib/live_debugger_refactor/services/callback_tracer/gen_servers/tracing_manager.ex
```diff
@@ -29,32 +31,32 @@ defmodule LiveDebuggerRefactor.Services.CallbackTracer.GenServers.TracingManager
   def handle_info(:setup_tracing, state) do
     TracingActions.setup_tracing!()
 
-    {:noreply, state}
+    noreply(state)
```
kraleppa, line 34: Imo when you don't pipe you should use standard convention `{:noreply, state}`. These are only helpers to make piping looks nicer

PR #603, opened 2025-07-16
Refactor: Create `ProcessMonitor` genserver

hhubert6:


lib/live_debugger_refactor/services/process_monitor/actions.ex
```diff
@@ -0,0 +1,54 @@
...
+  alias LiveDebuggerRefactor.Services.ProcessMonitor.GenServers.ProcessMonitor
+
+  alias LiveDebuggerRefactor.Services.ProcessMonitor.Events.LiveViewBorn
+  alias LiveDebuggerRefactor.Services.ProcessMonitor.Events.LiveViewDied
+  alias LiveDebuggerRefactor.Services.ProcessMonitor.Events.ComponentCreated
+  alias LiveDebuggerRefactor.Services.ProcessMonitor.Events.ComponentDeleted
+
+  @spec register_component_created(ProcessMonitor.state(), pid(), CommonTypes.cid()) ::
+          ProcessMonitor.state()
+  def register_component_created(state, pid, cid) do
```
kraleppa, line 18: I think it can raise an exception so you should use `!`
```suggestion
  def register_component_created!(state, pid, cid) do
```

Similarly in below

lib/live_debugger_refactor/services/process_monitor/gen_servers/process_monitor.ex
```diff
@@ -0,0 +1,92 @@
...
+
+  def start_link(opts \\ []) do
+    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+  end
+
+  @impl true
+  def init(_opts) do
+    Bus.receive_traces!()
+
+    ok(%{})
```
kraleppa, line 28: I mean this helper is useful but in such case I'd go the classic way
```suggestion
    {:ok, %{}}
```

lib/live_debugger_refactor/services/callback_tracer/events.ex
```diff
@@ -0,0 +1,23 @@
...
+  @moduledoc """
+  Temporary events for LiveDebuggerRefactor.Services.CallbackTracer.
+  """
+
+  use LiveDebuggerRefactor.Event
+
+  alias LiveDebugger.CommonTypes
+
+  defevent(TraceCalled,
+    id: neg_integer(),
```
kraleppa, line 11: I'd be more specific here because it looks like id of event, not id of trace. Also I think it may be `0` actually, so it should be `non_positive_integer()` 🤔 
```suggestion
    trace_id: neg_integer(),
```

lib/live_debugger_refactor/services/process_monitor/gen_servers/process_monitor.ex
```diff
@@ -0,0 +1,92 @@
...
+    |> noreply()
+  end
+
+  @impl true
+  def handle_info(
+        %TraceCalled{
+          module: Phoenix.LiveView.Diff,
+          function: :delete_component,
+          cid: cid,
+          context: %{pid: pid}
```
kraleppa, line 53: I'm wondering actually if `pid` in such case shouldn't be a part of actual event and context 🤔 Not sure if context is useful now when we decided to store more inside events... Maybe we could keep there info about broadcaster of this event, like its PID or broadcaster module

lib/live_debugger_refactor/services/process_monitor/gen_servers/process_monitor.ex
```diff
@@ -0,0 +1,92 @@
...
+  def handle_info(_, state) do
+    noreply(state)
+  end
+
+  defp maybe_register_component_created(state, pid, cid) do
+    if MapSet.member?(state[pid], cid) do
+      state
+    else
+      state
+      |> Actions.register_component_created(pid, cid)
```
kraleppa, line 80: I'd flatten single pipes

lib/live_debugger_refactor/services/process_monitor/gen_servers/process_monitor.ex
```diff
@@ -0,0 +1,92 @@
...
+  This module is monitoring the status of LiveView processes created by a debugged application.
+  """
+
+  use GenServer
+
+  alias LiveDebuggerRefactor.CommonTypes
+  alias LiveDebuggerRefactor.Bus
+  alias LiveDebuggerRefactor.Services.ProcessMonitor.Actions
+  alias LiveDebuggerRefactor.Services.CallbackTracer.Events.TraceCalled
+  alias LiveDebuggerRefactor.Services.CallbackTracer.Events.TraceReturned
```
kraleppa, line 12: Could you make some order in these aliases - I'd like bus and events to be close together separated from other aliases
```suggestion
  alias LiveDebuggerRefactor.CommonTypes
  alias LiveDebuggerRefactor.Services.ProcessMonitor.Actions
  
  alias LiveDebuggerRefactor.Bus
  alias LiveDebuggerRefactor.Services.CallbackTracer.Events.TraceCalled
  alias LiveDebuggerRefactor.Services.CallbackTracer.Events.TraceReturned
```

kraleppa: Also, I'd consider aliasing `ProcessMonitor.Actions` as `ProcessMonitorActions` like you did in tests

lib/live_debugger_refactor/services/process_monitor/gen_servers/process_monitor.ex
```diff
@@ -0,0 +1,92 @@
+defmodule LiveDebuggerRefactor.Services.ProcessMonitor.GenServers.ProcessMonitor do
```
GuzekAlan, line 1: It still feels weird and I know I've done same convention but there must be a better way to name it. How about just not nesting it in `gen_servers` but make it `.../sercices/process_monitor/gen_server.ex` ? 

cc @kraleppa

lib/live_debugger_refactor/services/process_monitor/actions.ex
```diff
@@ -0,0 +1,54 @@
...
+  alias LiveDebuggerRefactor.Services.ProcessMonitor.Events.LiveViewBorn
+  alias LiveDebuggerRefactor.Services.ProcessMonitor.Events.LiveViewDied
+  alias LiveDebuggerRefactor.Services.ProcessMonitor.Events.LiveComponentCreated
+  alias LiveDebuggerRefactor.Services.ProcessMonitor.Events.LiveComponentDeleted
+
+  @spec register_component_created!(ProcessMonitor.state(), pid(), CommonTypes.cid()) ::
+          ProcessMonitor.state()
+  def register_component_created!(state, pid, cid) do
+    new_state = Map.update!(state, pid, &MapSet.put(&1, cid))
+    Bus.broadcast_event!(%LiveComponentCreated{cid: cid}, pid)
```
GuzekAlan, line 20: Shouldn't we broadcast on both `lvdbg/*` and `lvdbg/PID` ? We talked about adding braodcast when you need to subscribe to it

hhubert6: That is exactly how `Bus.broadcast_event!/2` works

lib/live_debugger_refactor/common_types.ex
```diff
@@ -0,0 +1,7 @@
+defmodule LiveDebuggerRefactor.CommonTypes do
```
GuzekAlan, line 1: I know it is not in scope of this task but could you update alias in `TraceStorage` for this one please?

lib/live_debugger_refactor/services/helpers.ex
```diff
@@ -0,0 +1,11 @@
+defmodule LiveDebuggerRefactor.Services.Helpers do
+  @moduledoc false
+
+  @type state() :: term()
+
+  @spec ok(state()) :: {:ok, state()}
```
kraleppa, line 6: Why `term()` here is not ok?

kraleppa (approved):

PR #615, opened 2025-07-22
Refactor: send event after state save

GuzekAlan:


lib/live_debugger_refactor/services/state_manager/actions.ex
```diff
@@ -14,6 +17,7 @@ defmodule LiveDebuggerRefactor.Services.StateManager.Actions do
   def save_state!(pid) when is_pid(pid) do
     with {:ok, lv_state} <- LiveViewDebug.liveview_state(pid) do
       StatesStorage.save!(lv_state)
+      Bus.broadcast_state!(%StateChanged{pid: pid})
```
hhubert6, line 20:
```suggestion
      Bus.broadcast_state!(%StateChanged{pid: pid}, pid)
```

hhubert6 (approved):

lib/live_debugger_refactor/services/state_manager/actions.ex
GuzekAlan: Hmm. I think that `broadcast_trace!/1` and `broadcast_state!/1` won't be used.

PR #616, opened 2025-07-23
Refactor: Create `Discovery` context

hhubert6:


lib/live_debugger_refactor/app/utils/parsers.ex
```diff
@@ -0,0 +1,37 @@
...
+  @spec string_to_pid(string :: String.t()) :: {:ok, pid()} | :error
+  def string_to_pid(string) when is_binary(string) do
+    if String.match?(string, ~r/[0-9]+\.[0-9]+\.[0-9]+/) do
+      {:ok, :erlang.list_to_pid(~c"<#{string}>")}
+    else
+      :error
+    end
+  end
+
+  @spec cid_to_string(cid :: struct()) :: String.t()
```
GuzekAlan, line 26: there is type for `cid`

lib/live_debugger_refactor/app/utils/parsers.ex
```diff
@@ -0,0 +1,37 @@
...
+  @spec pid_to_string(pid()) :: String.t()
+  def pid_to_string(pid) when is_pid(pid) do
+    pid
+    |> :erlang.pid_to_list()
+    |> to_string()
+    |> String.slice(1..-2//1)
+  end
+
+  @spec string_to_pid(string :: String.t()) :: {:ok, pid()} | :error
+  def string_to_pid(string) when is_binary(string) do
```
GuzekAlan, line 18: could you add a doc with format of string which is used to convert?

GuzekAlan (approved):

lib/live_debugger_refactor/app/discovery/web/window_discovery_live.ex
kraleppa, line 1: We discussed this with @GuzekAlan a while ago that this view could be just a variant of `discovery_live` based on live_actions. I'm wondering if we shouldn't perhaps merge these two views now when we're refactoring and they're not fully implemented and quite small

GuzekAlan: Yesterday we've come to agreement that it will be better to leave it as it to avoid conditional renders and conditional logic. Also the similar logic (the argument for why to do it in one LiveView as in `debugger`) won't be very big. We will see if this needs a refactor after new live view discovery

kraleppa: ok, we can proceed this way - thanks for clarification

kraleppa (approved):

PR #623, opened 2025-07-25
Feature: add menu to debug button

kraleppa:
Decided that entering inspect mode should be started from the browser to simulate a flow of right clicking on element and choosing `inspect` in traditional DevTools.

Since we need a way to start inspect mode - I decided to extend debug button. I have a feeling that its code become a bit clumsy, but not sure how to improve it using pure JS.

https://github.com/user-attachments/assets/90ee7def-6a82-43cc-8466-fd2e1d969e8e


assets/js/client/debug_panel/debug_menu.js
```diff
@@ -0,0 +1,63 @@
...
+    <div id="debug-tooltip" style="
+      position: absolute;
+      background: white;
+      border: 1px solid #e2e8f0;
+      border-radius: 8px;
+      box-shadow: 0 10px 25px rgba(0, 0, 0, 0.15);
+      min-width: 160px;
+      z-index: 10000;
+      display: none;
+      font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, sans-serif;
```
GuzekAlan, line 12: does this font work on other system than MacOS?

assets/js/client/debug_panel.js
```diff
@@ -0,0 +1,147 @@
...
+      debugButton.style.top = `${newTop}px`;
+      debugButton.style.right = 'auto';
+      debugButton.style.bottom = 'auto';
+    }
+  };
+
+  const showMenu = () => {
+    const buttonRect = debugButton.getBoundingClientRect();
+    const menuWidth = 160;
+    const menuHeight = 120;
```
GuzekAlan, line 71: these are magic values. The width is `min-w` in `debugMenu` so it might not change accordingly when text in menu will change

GuzekAlan (commented):
Looks way better now. Good job!

PR #624, opened 2025-07-28
Refactor: Create `node_state` context (part II)

hhubert6:


lib/live_debugger_refactor/app/utils/term_parser.ex
```diff
@@ -4,14 +4,36 @@ defmodule LiveDebuggerRefactor.App.Utils.TermParser do
...
+    - `children`: A list of child nodes, if any.
+    - `content`: Display elements that represent the content of the node when has no children or not expanded.
+    - `expanded_before`: Display elements shown before the node's children when expanded.
+    - `expanded_after`: Display elements shown after the node's children when expanded.
+    """
+    defstruct [:kind, :children, :content, :expanded_before, :expanded_after]
+
+    @type t :: %__MODULE__{
+            kind: String.t(),
+            children: [TermNode.t()] | nil,
```
kraleppa, line 31: Wouldn't it make more sense if these values could be just list and if there are no `TermNode` the list is empty?

kraleppa (approved):
Nice!

PR #637, opened 2025-07-29
Refactor: Add HookComponents for `callback_tracing`

GuzekAlan:
I've added behaviour for HookComponents, and moved the components to compose a trace into `Components`. Filters will be done in other PR since this one is already big enough

### DIFF
--- a/lib/live_debugger_refactor/app/debugger/callback_tracing/trace_display.ex
+++ b/lib/live_debugger_refactor/app/debugger/callback_tracing/trace_display.ex
@@ -1,22 +1,27 @@
 defmodule LiveDebuggerRefactor.App.Debugger.CallbackTracing.TraceDisplay do
   @moduledoc """
   Wrapper for trace to display it in the UI.
+
+  * `id` - id of the trace
+  * `trace` - trace to display
+  * `from_event?` - whether the trace comes from an event
+  * `render_body?` - whether to render the body of the trace
   """
 
   alias LiveDebuggerRefactor.Structs.Trace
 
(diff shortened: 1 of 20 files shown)


lib/live_debugger_refactor/app/debugger/callback_tracing/trace_display.ex
kraleppa, line 1: maybe `structs` directory? 👀

lib/live_debugger_refactor/app/debugger/callback_tracing/trace_display.ex
```diff
@@ -0,0 +1,26 @@
...
+  """
+
+  alias LiveDebuggerRefactor.Structs.Trace
+
+  defstruct [:id, :trace, :from_tracing?, :counter, render_body?: false]
+
+  @type t() :: %__MODULE__{
+          id: integer(),
+          trace: Trace.t(),
+          from_tracing?: boolean(),
```
kraleppa, line 13: This name `from_tracing?` is quite unintuitive for me... Maybe some comment or renaming it?

kraleppa (approved):
Nice!

lib/live_debugger_refactor/app/debugger/callback_tracing/trace_display.ex
GuzekAlan: It might actually change a bit since now we store in ETS and don't send whole trace via pubsub. So I assume that it will be something like "from_event?". But I will change the workign and creating TraceDisplay when connecting to events ok?

hhubert6 (approved):
LGTM

PR #653, opened 2025-08-05
Refactor: Add nested LiveViews and missing components

GuzekAlan:
I've added temporary commit to make the components visible and clickable (at least partially). I will delete after review


hhubert6 (approved):
👍

PR #665, opened 2025-08-11
Refactor: Add DebuggerLive, small fixes, add missing modules

GuzekAlan:


lib/live_debugger_refactor/app/debugger/queries/lv_process.ex
```diff
@@ -0,0 +1,44 @@
...
+
+  def parent_lv_process(_lv_process), do: nil
+
+  @spec fetch_with_retries(pid()) :: LvProcess.t() | nil
+  def fetch_with_retries(pid) when is_pid(pid) do
+    with nil <- fetch_after(pid, 200),
+         nil <- fetch_after(pid, 800) do
+      fetch_after(pid, 1000)
+    end
+  end
```
hhubert6, line 24: It it really necessary to fetch `LvProcess` this way? Is there a common scenario in which it helps to load it after delay?

lib/live_debugger_refactor/app/debugger/queries/node.ex
```diff
@@ -0,0 +1,34 @@
...
+  alias LiveDebuggerRefactor.API.StatesStorage
+
+  @spec module_from_id(TreeNode.id(), pid()) :: {:ok, module()} | :error
+  def module_from_id(node_id, pid) when is_node_id(node_id) do
+    with %LvState{} = state <- StatesStorage.get!(pid),
+         {:ok, %TreeNode{module: module}} <- get_node(state, node_id) do
+      {:ok, module}
+    else
+      _ -> :error
+    end
```
hhubert6, line 19: What about a situation where there's no state in storage? In case of `LvProcess` you fetch the socket directly from LiveView process. You're not doing it here on purpose?

lib/live_debugger_refactor/app/debugger/events.ex
```diff
@@ -0,0 +1,11 @@
+defmodule LiveDebuggerRefactor.App.Debugger.Events do
+  @moduledoc """
+  Events broadcasted by the Debugger context.
+  """
+
+  use LiveDebuggerRefactor.Event
+
+  alias LiveDebuggerRefactor.App.Debugger.Structs.TreeNode
+
+  defevent(NodeIdParamChanged, node_id: TreeNode.id(), debugger_pid: pid())
```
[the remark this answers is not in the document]
GuzekAlan: As discussed IRL I would stick to the "params groups" being send rather then all params (e. g. sending all filters params in one event). We could also send all params but make this event more structured (to avoid passing all params from `query_string` to all the NestedLiveViews. It might also make it easier to send events only when params changed

hhubert6 (commented):
What about `DebuggerMounted` and `DebuggerTerminated` events? You plan to do it in a separate PR?

kraleppa (approved):

PR #670, opened 2025-08-11
Refactor: Add `DisplayNewTraces` hook

hhubert6:
I tried to mitigate this unpleasant effect when filtering by execution time. When callback takes at least a few hundred milliseconds after call trace is displayed with timer and then is removed if filter does not match. However when callback is quick (something around under 10ms) it sometimes is not displayed at all and sometimes it shows for a split second. With many callbacks running traces blinks which randomly. I tried to resolve this with some debouncing.

### DIFF
--- a/lib/live_debugger_refactor/app/debugger/callback_tracing/web/hooks/display_new_traces.ex
+++ b/lib/live_debugger_refactor/app/debugger/callback_tracing/web/hooks/display_new_traces.ex
@@ -1,7 +1,11 @@
 defmodule LiveDebuggerRefactor.App.Debugger.CallbackTracing.Web.Hooks.DisplayNewTraces do
   @moduledoc """
   This hook is responsible for displaying new traces.
-  It is used to display new traces when the user starts tracing.
+  It is used to display new traces and filter them by execution time when the user starts tracing.
+
+  It incorporates debouncing `TraceCalled` event. It helps to reduce inserting and deleting traces from
+  a stream in a very short intervals (e.g. when callback execution time is only couple of microseconds
+  and they don't match the filter they appear in the UI and are removed very quickly which casuses blinking effect).
   """
 
   use LiveDebuggerRefactor.App.Web, :hook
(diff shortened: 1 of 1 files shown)


lib/live_debugger_refactor/app/debugger/callback_tracing/web/hooks/display_new_traces.ex
```diff
@@ -0,0 +1,121 @@
...
+  alias LiveDebuggerRefactor.Structs.Trace
+  alias LiveDebuggerRefactor.API.TracesStorage
+  alias LiveDebuggerRefactor.App.Debugger.CallbackTracing.Web.Helpers.Filters, as: FiltersHelpers
+  alias LiveDebuggerRefactor.App.Debugger.CallbackTracing.Structs.TraceDisplay
+
+  alias LiveDebuggerRefactor.Services.CallbackTracer.Events.TraceCalled
+  alias LiveDebuggerRefactor.Services.CallbackTracer.Events.TraceReturned
+  alias LiveDebuggerRefactor.Services.CallbackTracer.Events.TraceErrored
+
+  @debounce_timeout 1
```
GuzekAlan, line 18: could you add unit of timeout to the name?

lib/live_debugger_refactor/app/debugger/callback_tracing/web/hooks/display_new_traces.ex
```diff
@@ -0,0 +1,121 @@
+defmodule LiveDebuggerRefactor.App.Debugger.CallbackTracing.Web.Hooks.DisplayNewTraces do
+  @moduledoc """
+  This hook is responsible for displaying new traces.
```
GuzekAlan, line 3: maybe add comment on why there is a `debounce` event. I feel that we might forget about it later

lib/live_debugger_refactor/app/debugger/callback_tracing/web/hooks/display_new_traces.ex
```diff
@@ -0,0 +1,121 @@
...
+  Initializes the hook by attaching the hook to the socket and checking the required assigns, other hooks and streams.
+  """
+  @spec init(Phoenix.LiveView.Socket.t()) :: Phoenix.LiveView.Socket.t()
+  def init(socket) do
+    socket
+    |> check_hook!(:tracing_fuse)
+    |> check_assigns!(@required_assigns)
+    |> check_stream!(:existing_traces)
+    |> check_private!(:live_stream_limit)
+    |> put_private(:canceled, false)
```
GuzekAlan, line 35: could you add more descriptive name

kraleppa (approved):
Nice idea with using debounce!

GuzekAlan (approved):

PR #679, opened 2025-08-13
Feature: display node info during highlighting

kraleppa:
https://github.com/user-attachments/assets/acb93df7-9ec1-4dca-8a68-11a9d2b53034

### DIFF
--- a/assets/client/services/inspect.js
+++ b/assets/client/services/inspect.js
@@ -89,6 +89,11 @@ export default function initElementInspection({
     }
   };
 
+  const handleMouseLeave = () => {
+    pushClearEvent();
+    lastID = null;
+  };
+
   const disableInspectMode = () => {
     if (!inspectMode) {
       return;
@@ -107,6 +112,7 @@ export default function initElementInspection({
     document.body.classList.remove('live-debugger-inspect-mode');
     document.body.removeEventListener('click', handleInspect);
     document.body.removeEventListener('mouseover', handleMove);
+    document.removeEventListener('mouseleave', handleMouseLeave);
     document.removeEventListener('contextmenu', handleRightClick);
     document.removeEventListener('keydown', handleEscape);
   };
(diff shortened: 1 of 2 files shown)


assets/client/services/highlight.js
```diff
@@ -25,7 +27,8 @@ function createHighlightElement(activeElement, detail, id) {
   highlight.style.left = `${rect.left + window.scrollX}px`;
   highlight.style.width = `${activeElement.offsetWidth}px`;
   highlight.style.height = `${activeElement.offsetHeight}px`;
-  highlight.style.backgroundColor = '#87CCE880';
+  highlight.style.backgroundColor =
+    detail.type === 'LiveComponent' ? '#87CCE880' : '#b1dfd080';
```
hhubert6, line 31: I'd make these colors more distinguishable

assets/client/client.js
```diff
@@ -30,6 +32,10 @@ window.document.addEventListener('DOMContentLoaded', function () {
         console.log('Received ping', resp);
         debugChannel.push('pong', resp);
       });
+
+      initWindowIdentifier(debugChannel, socketID);
```
hhubert6, line 36: Didn't you change the name of this function to `pushWindowInitialized` and moved it to `initDebugSocket`?

lib/live_debugger_refactor/app/debugger/components_tree/web/components_tree_live.ex
```diff
@@ -162,9 +162,18 @@ defmodule LiveDebuggerRefactor.App.Debugger.ComponentsTree.Web.ComponentsTreeLiv
 
   defp highlight_element(
          %{assigns: %{highlight?: true, lv_process: %{pid: pid}}} = socket,
-         %{"search-attribute" => attr, "search-value" => val}
+         %{"search-attribute" => attr, "search-value" => val} = params
```
hhubert6, line 165: Why don't you just do it like this? or use `params` for everything
```suggestion
         %{
           "search-attribute" => attr,
           "search-value" => val,
           "module" => module,
           "type" => type,
           "id" => id
         }

```

hhubert6 (commented):
🔥 Looks amazing 🔥

What do you think about adding the same icon as in components tree next to the module name in the tooltip when inspecting?

assets/client/client.js
kraleppa: Yeah, I did - it probably made mistake when merging main to this branch

PR #680, opened 2025-08-14
clean up tests

GuzekAlan:


lib/live_debugger_refactor/app/debugger/callback_tracing/web/hook_components/filters_fullscreen.ex
```diff
@@ -110,6 +110,7 @@ defmodule LiveDebuggerRefactor.App.Debugger.CallbackTracing.Web.HookComponents.F
   defp handle_event("reset-filters", _, socket) do
     socket
     |> assign(:current_filters, FiltersHelpers.default_filters(socket.assigns.node_id))
+    |> Hooks.ExistingTraces.assign_async_existing_traces()
```
GuzekAlan, line 113: this fixes reset filter not working

lib/live_debugger_refactor/app/debugger/callback_tracing/web/hooks/filter_new_traces.ex
```diff
@@ -51,10 +52,15 @@ defmodule LiveDebuggerRefactor.App.Debugger.CallbackTracing.Web.Hooks.FilterNewT
...
-      %Phoenix.LiveComponent.CID{} = cid -> cid == trace_event.cid
+    case socket.assigns.node_id do
+      nil ->
+        true
+
+      pid when is_pid(pid) ->
+        pid == trace_event.pid && trace_event.cid == nil
+
+      %Phoenix.LiveComponent.CID{} = cid ->
+        cid == trace_event.cid
```
GuzekAlan, line 63: This fixes LiveComponent traces shown when nod_id is pid

lib/live_debugger_refactor/app/debugger/callback_tracing/web/node_traces_live.ex
```diff
@@ -138,4 +145,17 @@ defmodule LiveDebuggerRefactor.App.Debugger.CallbackTracing.Web.NodeTracesLive d
...
+  def handle_info(
+        %NodeIdParamChanged{node_id: node_id, debugger_pid: pid},
+        %{assigns: %{parent_pid: pid}} = socket
+      ) do
+    socket
+    |> assign(:node_id, node_id)
+    |> FiltersAssigns.assign_current_filters()
+    |> Hooks.TracingFuse.disable_tracing()
+    |> Hooks.ExistingTraces.assign_async_existing_traces()
+    |> noreply()
```
GuzekAlan, line 159: This fixes traces not responding to NodeChangged

lib/live_debugger_refactor/services/callback_tracer/gen_servers/tracing_manager.ex
```diff
@@ -17,6 +17,13 @@ defmodule LiveDebuggerRefactor.Services.CallbackTracer.GenServers.TracingManager
     GenServer.start_link(__MODULE__, opts, name: __MODULE__)
   end
 
+  @doc """
+  Checks if GenServer has been loaded
+  """
+  def ping!() do
+    GenServer.call(__MODULE__, :ping)
+  end
```
GuzekAlan, line 25: It is for E2E tests

lib/live_debugger_refactor/app/debugger/callback_tracing/web/components/filters.ex
```diff
@@ -3,7 +3,7 @@ defmodule LiveDebuggerRefactor.App.Debugger.CallbackTracing.Web.Components.Filte
   UI components for filters form.
   """
 
-  use LiveDebuggerWeb, :component
+  use LiveDebuggerRefactor.App.Web, :component
```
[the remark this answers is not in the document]
GuzekAlan: Even `Phoenix.LiveDashboard` is doing `.Web` so I think that this convention is not something which we must stick to. If we were to use it then we should do it by adding `app_web.ex` file and moving only rendering to web and data loading, and actions to non web. But I think it would be an overkill at this point

lib/live_debugger_refactor/app/debugger/callback_tracing/web/hooks/filter_new_traces.ex
```diff
@@ -12,7 +12,8 @@ defmodule LiveDebuggerRefactor.App.Debugger.CallbackTracing.Web.Hooks.FilterNewT
   alias LiveDebuggerRefactor.Services.CallbackTracer.Events.TraceErrored
 
   @required_assigns [
-    :current_filters
+    :current_filters,
+    :node_id
```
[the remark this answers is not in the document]
GuzekAlan: `GlobalCallback` is setting `node_id` to `nil`. It won't create a behaviour in which we have `nil` in `NodeInspector` and everything "works" because we have `Map.get(assigns, :node_id, nil)`. So that's why `node_id` is being required by this hook

PR #694, opened 2025-08-21
Bug: fix highlighting on dead view mode

GuzekAlan:


kraleppa (commented):
Is there a chance to create test for this?

kraleppa:
@GuzekAlan please do it - I think it is a bit misleading when you can inspect dead LiveView

PR #697, opened 2025-08-21
Refactor: Use DebugSocket in components highlighting

hhubert6:


kraleppa (approved):

assets/client/services/highlight.js
```diff
@@ -160,12 +169,13 @@ function showTooltip(detail) {
   dispatchCustomEvent('lvdbg:show-tooltip', props);
 }
 
-export default function initHighlight() {
+export default function initHighlight(debugChannel) {
   document.addEventListener('lvdbg:inspect-highlight', handleHighlight);
   document.addEventListener('lvdbg:inspect-pulse', handlePulse);
   document.addEventListener('lvdbg:inspect-clear', removeHighlightElement);
```
GuzekAlan, line 175: Why these are not handled by `debugChannel` ?

GuzekAlan: Hmm then shouldn't we also make inspecting via `debugChannel`? It can be done in other PR but I'm just curious if this was meant to be done by `EventListeners`

cc @kraleppa

GuzekAlan (approved):

assets/client/services/highlight.js
kraleppa: Yes, these events are triggered when you move your cursor when you are in the inspect mode or (in case of pulse) when you click on given element when in inspect mode. They're sent from the browser to the browser

PR #727, opened 2025-09-01
Enhancement: Add :info flash

srzeszut:


assets/app/styles/themes/dark.css
```diff
@@ -71,5 +71,11 @@
 
     --search-highlight-bg: var(--swm-yellow-100);
     --search-highlight-text: var(--gray-900);
+
+    /* Info */
+    --info-bg: var(--surface-0-bg);
```
kraleppa, line 76: I think in designs there is `bg-primary`. Did you change it on purpose?

hhubert6 (commented):
~~For me it blends too much with the website. Maybe you could make it slightly green or blue?~~

assets/app/styles/themes/dark.css
srzeszut: I think this is the same color. We don't have bg-primary and when i was checking the design it was surface-0-bg as the background

hhubert6 (approved):

PR #731, opened 2025-09-03
Bug: Checking if module has `:module_info` exported

kraleppa:
~~This is a draft since I couldn't reproduce it, but based on the error provided in issue it should do the trick~~


hhubert6 (approved):
nice

PR #732, opened 2025-09-05
Bug: Lack of exception trace

hhubert6:
This behavior happen due to reading state of dead live view (while saving it) 
```elixir
:ok <- StateActions.maybe_save_state!(updated_trace),
```

which stops trace handling, raises error and crashes gen_server. When gen_server is started again all other traces that were stored in a gen_server inbox are lost and thus I assume the handle_info which ended with exception with them.

```elixir
def handle_cast({:new_trace, {_, pid, type, {module, fun, _}, _, return_ts}, _n}, state)
    when fun in @allowed_callbacks and type in [:return_from, :exception_from] do
  with trace_key <- {pid, module, fun},
       {ref, trace, ts} <- get_trace_record(state, trace_key),
       execution_time <- calculate_execution_time(return_ts, ts),
       params <- %{execution_time: execution_time, type: type},
       {:ok, updated_trace} <- TraceActions.update_trace(trace, params),
       {:ok, ref} <- TraceActions.persist_trace(updated_trace, ref),
       :ok <- StateActions.maybe_save_state!(updated_trace),
       :ok <- TraceActions.publish_trace(updated_trace, ref) do
    {:noreply, delete_trace_record(state, trace_key)}
  else
    :trace_record_not_found ->
      {:noreply, state}

    {:error, err} ->
      raise "Error while handling trace: #{inspect(err)}"
  end
end
```

Simplest solution is to keep tracing flow despite of the result of saving state
```elixir
_ <- StateActions.maybe_save_state!(updated_trace),
```

### DIFF
--- a/test/e2e/dead_view_mode_test.exs
+++ b/test/e2e/dead_view_mode_test.exs
@@ -45,6 +45,31 @@ defmodule LiveDebugger.E2E.DeadViewModeTest do
     |> assert_has(css("label.pointer-events-none", text: "Highlight"))
   end
 
+  @sessions 2
+  feature "traces ended with exception are visible in dead view mode", %{
+    sessions: [dev_app, debugger]
+  } do
+    LiveDebugger.Services.CallbackTracer.GenServers.TracingManager.ping!()
+    LiveDebugger.API.SettingsStorage.save(:dead_view_mode, true)
+
+    dev_app
+    |> visit(@dev_app_url)
+
+    debugger
+    |> visit("/")
+    |> click(first_link())
+    |> click(global_callback_traces_button())
+    |> click(clear_traces_button())
+    |> click(toggle_tracing_button())
+
+    dev_app
+    |> click(css("button[phx-click=\"crash\"]"))
+
+    debugger
+    |> find(global_traces(count: 1))
+    |> assert_has(css("summary.border-error-icon"))
+  end
+
   defp global_traces(opts), do: css("#global-traces-stream details", opts)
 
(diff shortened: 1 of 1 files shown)


lib/live_debugger/services/process_monitor/actions.ex
```diff
@@ -35,13 +35,18 @@ defmodule LiveDebugger.Services.ProcessMonitor.Actions do
   def register_live_view_born!(state, pid, transport_pid) do
     Process.monitor(pid)
 
-    {:ok, components} = LiveViewDebug.live_components(pid)
-    node_ids = Enum.map(components, &%Phoenix.LiveComponent.CID{cid: &1.cid})
-    new_state = Map.put(state, pid, MapSet.new(node_ids))
+    case LiveViewDebug.live_components(pid) do
```
hhubert6, line 38: I also fixed handling of requesting dead live view here which sometimes happen

hhubert6:
> Looking good - is possible to test if exception trace does appear?

I've added a usual scenario of clicking crash button in LiveDebuggerDev.LiveViews.Main. Recreating the situation in which this exact bug occurred is very tedious so I didn't try to make it in test. Or do you insist to do it?

kraleppa:
@hhubert6 I think it's fine 

> > Looking good - is possible to test if exception trace does appear?
> 
> I've added a usual scenario of clicking crash button in LiveDebuggerDev.LiveViews.Main. Recreating the situation in which this exact bug occurred is very tedious so I didn't try to make it in test. Or do you insist to do it?
> 
> cc @kraleppa

I think it's good enough 👍

kraleppa (approved):

PR #734, opened 2025-09-08
Refactor: Add garbage collection to settings and remove config from settings storage

srzeszut:


lib/live_debugger/app/settings/web/settings_live.ex
```diff
@@ -109,6 +117,7 @@ defmodule LiveDebugger.App.Settings.Web.SettingsLive do
     |> case do
       {:ok, new_settings} ->
         assign(socket, settings: new_settings)
+        push_flash(socket, :info, "Setting updated successfully")
```
hhubert6, line 120: You need to pipe it to work properly
```suggestion
		socket
        |> assign(settings: new_settings)
        |> push_flash(:info, "Setting updated successfully")
```

lib/live_debugger/services/garbage_collector/gen_servers/garbage_collector.ex
```diff
@@ -4,13 +4,16 @@ defmodule LiveDebugger.Services.GarbageCollector.GenServers.GarbageCollector do
...
+  alias LiveDebugger.App.Events.UserChangedSettings
 
   alias LiveDebugger.Services.GarbageCollector.GenServers.TableWatcher
 
   alias LiveDebugger.Services.GarbageCollector.Actions.GarbageCollecting,
     as: GarbageCollectingActions
 
   alias LiveDebugger.Bus
+
   alias LiveDebugger.Services.GarbageCollector.Events.GarbageCollected
```
hhubert6, line 17: When aliasing try to group them logically so it's easier to read
```suggestion
  use GenServer
  
  alias LiveDebugger.API.SettingsStorage
  alias LiveDebugger.Services.GarbageCollector.GenServers.TableWatcher

  alias LiveDebugger.Services.GarbageCollector.Actions.GarbageCollecting,
    as: GarbageCollectingActions

  alias LiveDebugger.Bus
  alias LiveDebugger.Services.GarbageCollector.Events.GarbageCollected
  alias LiveDebugger.App.Events.UserChangedSettings
```

lib/live_debugger/services/garbage_collector/gen_servers/garbage_collector.ex
```diff
@@ -21,13 +24,18 @@ defmodule LiveDebugger.Services.GarbageCollector.GenServers.GarbageCollector do
...
-    init_garbage_collection_loop()
+    Bus.receive_events!()
+    garbage_collection_enabled? = SettingsStorage.get(:garbage_collection)
+    loop_garbage_collection()
 
-    {:ok, []}
+    {:ok,
+     %{
+       garbage_collection_enabled?: garbage_collection_enabled?
+     }}
```
hhubert6, line 34: Maybe like this?
```suggestion
    {:ok, %{garbage_collection_enabled?: SettingsStorage.get(:garbage_collection)}}
```

hhubert6 (commented):
Looks good

lib/live_debugger/app/settings/web/settings_live.ex
kraleppa: The funny thing it that I think he does not have to pipe it... 😄  `push_flash/3` does not change anything in the socket - it just  triggers `send(pid, {:put_flash, key, message})` where pid is the parent LiveView. The socket itself is not changed so this could theoretically work. 

But, I'd prefer if this pipe was actually added since we usually assume that this type of functions do change the socket. If we changed this function in the future with that assumption it would break

hhubert6: Right, but assigning new settings still should be last then because returned value should be new socket and because of that it's not working properly.

kraleppa (approved):

hhubert6 (approved):

PR #736, opened 2025-09-09
Bug: Fix appearance buttons in settings are not synchronized across tabs

srzeszut:


assets/app/app.js
```diff
@@ -75,6 +75,13 @@ function getCsrfToken() {
     .getAttribute('content');
 }
 
+this.handleStorage = (e) => {
+  if (e.key !== 'theme') return;
+  document.documentElement.classList.toggle('dark', e.newValue === 'dark');
+};
+
+window.addEventListener('storage', this.handleStorage);
```
hhubert6, line 83:
```suggestion
function handleStorage(e) {
  if (e.key !== 'theme') return;
  document.documentElement.classList.toggle('dark', e.newValue === 'dark');
};

window.addEventListener('storage', handleStorage);
```

hhubert6 (commented):
Nice

as I think of this it would be great to have e2e test for settings syncing across tabs (theme buttons and other switches)

PR #739, opened 2025-09-10
Add EndlessCrashLoop LiveView to devApp

srzeszut:


hhubert6 (commented):
The reason this LiveView is currently crashing is due to lack of `handle_async` (you can see it in the terminal logs). I think this example would better work if you would define `handle_async` and raised an error inside it. That way it can be nicely seen in the debugger that this handler crashed. 

Also in order for LiveDebugger to detect new LiveView is born it has to run `render` function. To simulate this scenario you need to give it some time, I'd add for example `Process.sleep(500)` inside function passed to `start_async`

PR #754, opened 2025-09-16
Explicit formats in  SocketDiscoveryController

rhcarvalho:
Fixes compilation warning.

Before:

```
==> live_debugger
Compiling 127 files (.ex)
warning: use LiveDebugger.App.Controllers.SocketDiscoveryController must receive the :formats option with the formats you intend to render. To keep compatibility within your app, you can list it as:

    formats: [html: "View", json: "View", ...]

Listing all formats your application renders.
```

After: no warning.

---

The controller only does a redirect, it never renders any format, so the empty list should be suitable for now.


srzeszut (approved):

srzeszut:
Thanks for your contribution!

Issue #526, opened 2025-06-26
Performance issues with assigns inspection


kraleppa:
We received an internal feedback that loading assigns sometimes takes too much time when they are very large. The problem is that we render the whole assigns structure at once with all nested levels. We may need to consider adding lazy loading - during the first render we render only 3 levels, and when user opens next level, we load another level asynchronusly.

You can talk to @p1003 since he reported this issue

Rutgerdj:
I'm also experiencing this issue. LiveDebugger is great for smaller projects, but the performance with large assigns is currently a dealbreaker for my main project.

PR #756, opened 2025-09-18
Feature: Better visual indication of assigns changes

hhubert6:
To keep things simple and don't interfere with other features (like searching in assigns) I think that visually indicating where particular value changed in assigns shouldn't open any collapsibles and it should be always turned on.
I tried to implement this approach here.

cc @GuzekAlan @srzeszut

### DIFF
--- a/lib/live_debugger/app/debugger/node_state/utils.ex
+++ b/lib/live_debugger/app/debugger/node_state/utils.ex
@@ -10,19 +10,23 @@ defmodule LiveDebugger.App.Debugger.NodeState.Utils do
   """
   @spec diff(map(), map()) :: map()
   def diff(map1, map2) when is_map(map1) and is_map(map2) do
-    all_keys = (Map.keys(map1) ++ Map.keys(map2)) |> Enum.uniq()
-
-    all_keys
+    (Map.keys(map1) ++ Map.keys(map2))
+    |> Enum.uniq()
     |> Enum.reduce(%{}, fn key, acc ->
-      v1 = Map.get(map1, key, :__missing__)
-      v2 = Map.get(map2, key, :__missing__)
+      result1 = Map.fetch(map1, key)
+      result2 = Map.fetch(map2, key)
 
-      case compare_values(v1, v2) do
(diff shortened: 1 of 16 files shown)


lib/live_debugger/app/debugger/node_state/utils.ex
```diff
@@ -0,0 +1,34 @@
...
+  leaf values are `true`.
+  """
+  @spec diff(map(), map()) :: map()
+  def diff(map1, map2) when is_map(map1) and is_map(map2) do
+    all_keys = (Map.keys(map1) ++ Map.keys(map2)) |> Enum.uniq()
+
+    all_keys
+    |> Enum.reduce(%{}, fn key, acc ->
+      v1 = Map.get(map1, key, :__missing__)
+      v2 = Map.get(map2, key, :__missing__)
```
GuzekAlan, line 18: Can you avoid using default value? I know probability that someone will have assign with value `:__missing__` is low but maybe using `Map.fetch/2` with `with` would do the trick?

lib/live_debugger/app/debugger/node_state/utils.ex
```diff
@@ -0,0 +1,34 @@
...
+  Utility functions for the Node State.
+  """
+
+  @doc """
+  Computes a recursive diff between two maps.
+  Returns a map of keys that changed, where
+  leaf values are `true`.
+  """
+  @spec diff(map(), map()) :: map()
+  def diff(map1, map2) when is_map(map1) and is_map(map2) do
```
GuzekAlan, line 12: Recursion here is not working for same keys in different nesting levels: 

Image

hhubert6: I don't entirely get what you mean here. Output on those two maps you provided is as expected. Could you maybe explain in more detail?

GuzekAlan (approved):
Looks nice 🦈

srzeszut (approved):

kraleppa (approved):

hhubert6:
Closing due to current major changes (#526) that must be done before this pr

PR #761, opened 2025-09-29
Enhancement: Add diffing functionality

GuzekAlan:


GuzekAlan:
This diffing functionality will be used to update `TermNode` and it could be used to save assigns history

PR #766, opened 2025-10-01
Enhancement: Extend `GarbageCollector`

hhubert6:


lib/live_debugger/services/callback_tracer/actions/state.ex
```diff
@@ -18,6 +19,11 @@ defmodule LiveDebugger.Services.CallbackTracer.Actions.State do
     do_save_state!(pid)
   end
 
+  def maybe_save_state!(%Trace{pid: pid, function: function, args: [_, _, socket]})
+      when function in [:mount, :handle_params] do
+    do_save_initial_state!(pid, socket)
```
kraleppa, line 24: I assume that this variant is used for LiveViews that dies immediately and we cannot fetch the state as usual via `LiveViewDebug.liveview_state/1` because it may be already dead, right? The problem with this approach is that you don't have a list of LiveComponents in the state storage which theoretically may affect LiveViews that do not die immediately.

What will happen if someone triggers `handle_params/3` that does not change any assigns? `handle_params/3` in this case will be the last callback triggered in given LiveView (`render/1` will not be triggered because assigns did not change). In such case components inspection may break

lib/live_debugger/services/garbage_collector/gen_servers/table_watcher.ex
```diff
@@ -78,22 +76,21 @@ defmodule LiveDebugger.Services.GarbageCollector.GenServers.TableWatcher do
...
+    Process.monitor(debugger_pid)
+
     state
     |> add_watcher(debugged_pid, debugger_pid)
     |> noreply()
   end
 
-  @impl true
-  def handle_info(%DebuggerTerminated{debugger_pid: debugger_pid}, state) do
+  def handle_info({:DOWN, _ref, :process, debugger_pid, _reason}, state) do
```
kraleppa, line 93: Why do we exactly change it? Using event is more flexible because you have to monitor in a single place only

lib/live_debugger/services/garbage_collector/gen_servers/garbage_collector.ex
```diff
@@ -37,39 +42,29 @@ defmodule LiveDebugger.Services.GarbageCollector.GenServers.GarbageCollector do
     watched_pids = TableWatcher.watched_pids()
     alive_pids = TableWatcher.alive_pids()
 
-    traces_collected = GarbageCollectingActions.garbage_collect_traces!(watched_pids, alive_pids)
-    states_collected = GarbageCollectingActions.garbage_collect_states!(watched_pids, alive_pids)
-
-    if traces_collected or states_collected do
-      Bus.broadcast_event!(%GarbageCollected{})
```
kraleppa, line 44: Why do we get rid of this event?

lib/live_debugger/services/garbage_collector/gen_servers/table_watcher.ex
hhubert6: `DebuggerTerminated` was sent in `terminate` callback but turns out for some reason it's not triggered every time. For example when returning from `debugger` to active LiveViews it sometimes wasn't called so the event wasn't sent and the state of `table_watcher' was wrong. Using `Process.monitor` and listening for `:DOWN` is more reliable in this case.

lib/live_debugger/services/garbage_collector/gen_servers/garbage_collector.ex
hhubert6: It wasn't used anywhere and so when I refactored this part of code I didn't try to keep if it's not even needed. We can always add it when there will be such a need.

lib/live_debugger/services/garbage_collector/gen_servers/table_watcher.ex
kraleppa: Ok, but isn't `process_monitor` service a better place for that? I'd move this `Process.monitor` there since we already have process monitoring there and keep `DebuggerTerminated` if we wanted to subscribe for it in the future

kraleppa: We created these events to make sure that the app itself is easily extensible in the future. We introduced this architecture patterns to make the future development easier for us and provide a better abstraction layer (you just subscribe to nice looking event instead monitoring the process in multiple places and handling it's messages). 

> I just think it's a bit confusing flow then that TableWatcher listens for DebuggerMounted as well as ProcessMonitor so it can listen for :DOWN and send it back to TableWatcher.

Not sure if get that. ProcessMonitor in my mind is something responsible for monitoring processes - it already monitors `debugged` processes, why can't it monitor also `debugger` processes? IMO it's a natural fit for this service and I think the only way why we decided not to put monitoring od `debugger` processes here from the beginning was that we wanted to rely on `terminate` callback (which happens to be unreliable).

IMO ProcessMonitor should be responsible for monitoring all LiveView processes, reacting to `:DOWN` event and broadcasting a corresponding event which can be received by TableWatcher

PR #785, opened 2025-10-06
Bug: Debugger LiveViews state saved among debugged LiveViews

hhubert6:


lib/live_debugger/services/callback_tracer/actions/trace.ex
GuzekAlan, line 42: returned type changed

GuzekAlan (approved):

PR #798, opened 2025-10-14
Feature: Add dead LiveViews section

hhubert6:


lib/live_debugger/app/debugger/callback_tracing/web/global_traces_live.ex
kraleppa: Shouldn't you handle `DeadViewModeEntered` event like you do in `node_traces_live.ex`?

lib/live_debugger/app/discovery/web/discovery_live.ex
kraleppa: I'd consider splitting dead and active liveviews section do LiveComponents

kraleppa:
> It might enhance the chance for someone to disable it and leave it which might cause memory related issues

I agree, but I'd probably try to increase the time dead LiveView is displayed - IMO current version where LiveViews are appearing and disappearing looks quite "buggy" - it gives a feeling that something is malfunctioning.

Also, since GC off is such undesirable state I'd rethink the banners with information about GC turned off and on that you created. I'd probably change the one which is displayed on GC off to `info` because this is not potentially dangerous situation, and thanks to that banner displayed on GC on will be more prominent 

cc @hhubert6

kraleppa (approved):

PR #799, opened 2025-10-15
Enhancement: Improve Assigns section performance

GuzekAlan:


hhubert6 (commented):
🐬 Amazing work 🐬

When assigns are really big the time required for render and update can take couple of seconds and It would be nice to have a loader indicating that the assigns are being updated

Also when testing on very big assigns html updates become increasingly longer but seems like it might be due to calculating assigns sizes (When I comment out displaying sizes it helps a lot). It's probably for another pr.

lib/live_debugger/app/debugger/node_state/web/hook_components/assigns_display.ex
```diff
@@ -0,0 +1,100 @@
...
+  end
+
+  attr(:id, :string, required: true)
+  attr(:node, TermNode, required: true)
+
+  @impl true
+  def render(assigns) do
+    ~H"""
+    <div id={@id}>
+      <.term node={@node} />
```
kraleppa: What's the point of having the whole HTML structure of this component extracted to the private function?

lib/live_debugger/app/debugger/web/components/elixir_display.ex
```diff
@@ -66,7 +61,7 @@ defmodule LiveDebugger.App.Debugger.Web.Components.ElixirDisplay do
 
   attr(:items, :list, required: true)
 
-  defp text_items(assigns) do
+  def text_items(assigns) do
```
kraleppa:
```suggestion
  defp text_items(assigns) do
```

Does it have to be public?

GuzekAlan:
@hhubert6 
> When assigns are really big the time required for render and update can take couple of seconds and It would be nice to have a loader indicating that the assigns are being updated

I think it is a good idea. What do you think about making gray overlay and not allowing user to interact when the UI is rendering? I would like to do i in separate PR since it's not crucial and some task are dependant on this one.


> Also when testing on very big assigns html updates become increasingly longer but seems like it might be due to calculating assigns sizes (When I comment out displaying sizes it helps a lot). It's probably for another pr.

I've added sending big messages to dev to better show this case. I'll create task for running assigns size calculations in the background

hhubert6 (approved):

srzeszut (approved):

PR #801, opened 2025-10-15
Feature: trace diffs sent to browser

kraleppa:
In this PR I introduced a new trace associated with LiveView diffs that are sent to the browser from LiveView via WebSocket. They can be accessed via global traces view, after selecting corresponding option in the filters form

Traces as a structure needs some unification, but since this PR is already big, I decided to refactor it in the next PR

https://github.com/user-attachments/assets/54180a8b-0750-4708-ae4e-086e097b6a74

### DIFF
--- a/test/e2e/global_traces_test.exs
+++ b/test/e2e/global_traces_test.exs
@@ -200,16 +200,10 @@ defmodule LiveDebugger.E2E.GlobalTracesTest do
     |> click(clear_traces_button())
     |> assert_has(traces(count: 0))
     |> assert_has(no_traces_info())
-
-    Process.sleep(100)
-
-    debugger
     |> assert_has(filters_sidebar())
-    |> click(trace_diffs_checkbox())
-
-    Process.sleep(200)
 
     debugger
+    |> click(trace_diffs_checkbox())
     |> click(apply_button())
     |> click(toggle_tracing_button())
 


lib/live_debugger/api/traces_storage.ex
```diff
@@ -163,10 +167,11 @@ defmodule LiveDebugger.API.TracesStorage do
     @traces_table_name :lvdbg_traces
     @processes_table_name :lvdbg_traces_processes
 
-    @type ets_elem() :: {integer(), Trace.t()}
+    @type ets_elem() :: {integer(), trace()}
     @type continuation() :: TracesStorage.continuation()
     @type ets_table_id() :: TracesStorage.ets_table_id()
     @type table_identifier() :: TracesStorage.table_identifier()
+    @type trace() :: Trace.t() | DiffTrace.t()
```
GuzekAlan: Why not use `LiveDebugger.API.TracesStorage.trace()` ?

lib/live_debugger/api/traces_storage.ex
```diff
@@ -449,31 +461,62 @@ defmodule LiveDebugger.API.TracesStorage do
 
       execution_times = Keyword.get(opts, :execution_times, %{})
       node_id = Keyword.get(opts, :node_id)
+      trace_diffs = Keyword.get(opts, :trace_diffs, false)
 
-      match_spec(node_id, functions, execution_times)
+      match_spec(node_id, functions, execution_times, trace_diffs)
```
GuzekAlan: Why not just do this


```suggestion
      match_spec(node_id, functions, execution_times)
      |> maybe_attach_diff_spec(trace_diffs)
```

kraleppa:
> Wondering if filter for seeing only diffs wouldn't be helpful.

Theoretically you can achieve that by filtering out every callback function. We can consider adding a button to form like "unselect all" - we'll be rethinking filters form anyway with the designer, so we can leave it for later

cc @hhubert6

PR #815, opened 2025-10-27
Refactor: unify trace data structures

kraleppa:
Decided not to change drastically our code - I just unified naming and types for these trace types.

The reason for such small changes is the fact that those types of traces are too different from each other and they have completely different flow in every place despite `TraceStorage`.

### DIFF
--- a/lib/live_debugger/app/debugger/callback_tracing/structs/trace_display.ex
+++ b/lib/live_debugger/app/debugger/callback_tracing/structs/trace_display.ex
@@ -73,7 +73,7 @@ defmodule LiveDebugger.App.Debugger.CallbackTracing.Structs.TraceDisplay do
   end
 
   defp get_type(%{type: :exception_from}), do: :error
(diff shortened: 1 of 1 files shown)


lib/live_debugger/app/debugger/callback_tracing/structs/trace_display.ex
```diff
@@ -71,13 +72,13 @@ defmodule LiveDebugger.App.Debugger.CallbackTracing.Structs.TraceDisplay do
   end
 
   defp get_type(%{type: :exception_from}), do: :error
-  defp get_type(%LiveDebugger.Structs.DiffTrace{}), do: :diff
+  defp get_type(%LiveDebugger.Structs.Trace.DiffTrace{}), do: :diff
```
srzeszut:
```suggestion
  defp get_type(%DiffTrace{}), do: :diff
```

srzeszut (approved):

PR #822, opened 2025-10-30
Enhancement: Add pulse on assign change

hhubert6:
Adapted from this poc #756


assets/app/hooks/diff_pulse.js
```diff
@@ -0,0 +1,31 @@
...
+
+      this.timeout = setTimeout(() => {
+        this.el.classList.remove('animate-diff-pulse');
+      }, 500);
+    }
+  },
+  destroyed() {
+    clearTimeout(this.timeout);
+  },
+};
```
kraleppa: Can't we do it like this? 
```suggestion
  mounted() {
    if (this.el.classList.contains("animate-diff-pulse")) {
      this.timeout = setTimeout(() => {
        this.el.classList.remove('animate-diff-pulse');
      }, 500);
    }
  },
  updated() {
    if (this.el.classList.contains("animate-diff-pulse")) {
      clearTimeout(this.timeout);
      
      this.timeout = setTimeout(() => {
        this.el.classList.remove('animate-diff-pulse');
      }, 500);
    }
  },
  destroyed() {
    clearTimeout(this.timeout);
  },
};
```

kraleppa (approved):

assets/app/hooks/diff_pulse.js
hhubert6: When assigns change in intervals shorter than the time of this animation it has to be reset manually to look properly. It might look confusing but this is what I came up with to achieve proper effect.

lib/live_debugger/app/debugger/node_state/web/hooks/node_assigns.ex
```diff
@@ -67,6 +70,8 @@ defmodule LiveDebugger.App.Debugger.NodeState.Web.Hooks.NodeAssigns do
     |> assign(:node_assigns_info, node_assigns_info)
     |> assign(:assigns_sizes, assigns_sizes)
     |> start_async(:fetch_node_assigns, fn ->
+      # Small sleep serves here as a debounce mechanism
+      Process.sleep(100)
```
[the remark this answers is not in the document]
hhubert6: Fetching is triggered by `StateChanged` which is broadcasted when any render happens in LiveView process (LiveComponents included). This leads to a race condition when second render (e.g. from LiveComponent) triggers fetching node assigns, which didn't change and it results in diff saying nothing changes. If this happens in a matter of milliseconds the effect is that pulse animation is overwritten and not visible. Adding short sleep makes it wait for the final render of this waterfall of renders.

PR #845, opened 2025-11-13
Feature: Add history of assigns

hhubert6:


lib/live_debugger/app/debugger/node_state/web/hook_components/assigns_history.ex
```diff
@@ -0,0 +1,266 @@
...
+      <span>No history records</span>
+    </div>
+    """
+  end
+
+  defp failed(assigns) do
+    ~H"""
+    <NodeStateComponents.failed />
+    """
+  end
```
kraleppa: Are those used? You use simple `:if` in the render and `<NodeStateComponents.failed />`

hhubert6: I've added this and forgot to swap in render, my bad.

PR #858, opened 2025-11-24
Enhancement: Update Active LiveViews page

hhubert6:
Designs introduce warning about GC being disabled on Active LiveViews page but I think having it on every page is a good idea. What do you think?

cc @kraleppa @srzeszut

### DIFF
--- a/test/e2e/dead_live_views_test.exs
+++ b/test/e2e/dead_live_views_test.exs
@@ -28,12 +28,10 @@ defmodule LiveDebugger.E2E.DeadLiveViewsTest do
 
     debugger
     |> visit("/")
-    |> assert_has(enable_dead_liveviews_checkbox(selected: false))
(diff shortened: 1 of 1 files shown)


kraleppa:
> Designs introduce warning about GC being disabled on Active LiveViews page but I think having it on every page is a good idea. What do you think?

I think it's a valid point - this warning is pretty universal and quite important, so displaying it on every page should be desired.

kraleppa:
This section seem to be jumping a bit when opened/closed (but it might be only in my mind 😅 )

Also it's missing a cursor pointer and I'm wondering if the title should be more gray when section is closed? 🤔 It's not on the designs and imo it does not make that much sense since the section seem entirely inactive, but the counter still changes...

https://github.com/user-attachments/assets/49eab058-c397-4f57-9d76-423ca6b021c6

PR #859, opened 2025-11-24
Refactor: Update layout of LiveDebugger

srzeszut:


lib/live_debugger/app/debugger/callback_tracing/web/global_traces_live.ex
```diff
@@ -100,76 +119,90 @@ defmodule LiveDebugger.App.Debugger.CallbackTracing.Web.GlobalTracesLive do
...
+        current_url={@url}
+        return_link={@return_link}
+      >
+        <:inspect_button>
+          <DebuggerHookComponents.InspectButton.render
+            inspect_mode?={@inspect_mode?}
+            lv_process={@lv_process}
+          />
+        </:inspect_button>
+      </NavigationMenu.sidebar>
```
hhubert6: If you're using navigation directly in `global_traces_live` then maybe do it like that in other pages instead of slots for better consistency.

hhubert6 (commented):
Sidebar broke a little. Close button is weirdly placed and report issue footer is too big.

<div style="display: flex">
Image
Image
</div>

Also content of the page in Node Inspector should have max-width but it takes full width.

Image

It's not directly concern of this pr but on Resources page on smaller screen widths sidebar button is displayed but there's no sidebar on this page. Also clicking it gives error in the console. As you already refactor navigation in this pr you can remove it too.

kraleppa:
Very nice, but I found some UI nitpicks:
1. This gaps should be the same
Screenshot 2025-11-27 at 11 08 55

2. Selected tab should be not only underlined, but also its color should change to swm blue

3. Tabs are jumping a bit when changing 
https://github.com/user-attachments/assets/fda7b740-2138-451c-b126-7d111ed21bd7

4. Add click-away to dropdown in mobile version of navigation bar

5. This copy button should be just next to the module name with some small gap (like gap-4)
Screenshot 2025-11-27 at 11 16 20

6. Disconnected state indicator should have the same design as connected state
Screenshot 2025-11-27 at 11 19 12

hhubert6 (approved):
👌

PR #878, opened 2025-12-11
Feature: Add highlighting active LiveViews

hhubert6:


lib/live_debugger/app/discovery/queries.ex
```diff
@@ -14,14 +14,14 @@ defmodule LiveDebugger.App.Discovery.Queries do
...
-  def fetch_grouped_lv_processes(transport_pid \\ nil) do
+  def fetch_grouped_lv_processes() do
     lv_processes =
-      with [] <- fetch_lv_processes_after(200, transport_pid),
-           [] <- fetch_lv_processes_after(800, transport_pid) do
-        fetch_lv_processes_after(1000, transport_pid)
+      with [] <- fetch_lv_processes_after(0),
+           [] <- fetch_lv_processes_after(500) do
+        fetch_lv_processes_after(1000)
       end
```
hhubert6: We no longer need to fetch processes for specific `transport_pid` (It was needed for `Active LiveViews in window` view). And initially fetching lv processes happened only when opening `Active LiveViews` and on manual refresh, that's why 200 ms delay was helpful for UX. But now we refresh on every event of `LiveViewBorn` and `LiveViewDied` so I think it's better not to delay anymore.

kraleppa (commented):
I think we need to disable highlighting for Dead LiveViews both in node inpsector and dead LiveViews section. It sometimes causes weird behavior when you can actually highlight components from LiveView that does not exist anymore...
Screenshot 2025-12-12 at 09 47 38

lib/live_debugger/app/debugger/components_tree/web/components_tree_live.ex
```diff
@@ -166,57 +147,49 @@ defmodule LiveDebugger.App.Debugger.ComponentsTree.Web.ComponentsTreeLive do
       ) do
     socket
     |> assign(highlight_disabled?: true)
```
kraleppa: Is this assign used anywhere? I think you removed it in before CR, and reverted after applying CR suggestions

kraleppa (approved):

Issue #481, opened 2025-06-06
Add validation and helpers to form


kraleppa:
The goal here is to make user who is filling send message form life easier. There is no precise scope in this task but creating as many validations and helpers for user is desired. Ideas:

Validating if provided code is a valid elixir code

Syntax highlighting

Autocompletion

Ideally textarea could be filled with example tuple after handler is chosen e.g.
When user chooses `handle_event` text area could be filled with
```Elixir
# {message, unassigned_params}
{"message", %{}}
```

PR #881, opened 2025-12-17
Feature: add form for sending user messages to node

kraleppa:
I haven't introduced any syntax highlighting or other quality of life features - it will be handled in #481 

https://github.com/user-attachments/assets/52d1812e-f83f-4932-ab63-6efe722e3af1

### DIFF
--- a/lib/live_debugger/app/debugger/web/live_components/send_event_fullscreen.ex
+++ b/lib/live_debugger/app/debugger/web/live_components/send_event_fullscreen.ex
@@ -72,15 +72,34 @@ defmodule LiveDebugger.App.Debugger.Web.LiveComponents.SendEventFullscreen do
                   <%= @message_error %>
                 </p>
               </div>
-              <.button variant="primary" type="submit" class="!w-full">
-                Send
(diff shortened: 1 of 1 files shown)


lib/live_debugger/app/debugger/web/live_components/node_basic_info.ex
```diff
@@ -40,31 +41,63 @@ defmodule LiveDebugger.App.Debugger.Web.LiveComponents.NodeBasicInfo do
...
     </div>
     """
   end
 
+  @impl true
+  def handle_event("open-send-event", _, socket) do
+    send_update(SendEventFullscreen,
+      id: "send-event-fullscreen",
+      reset_form?: true
+    )
```
hhubert6: I have a feeling that values in the form should stay even after closing and reopening for ease of use. It's kinda annoying it resets after every send. When someone's debugging a handler they might send same thing multiple times.

hhubert6 (commented):
Looks good

When module name is too long it overflows ugly
Image

PR #882, opened 2025-12-18
Enhacement: Remove `tracing_setup_delay`

hhubert6:
Solution involves setting up tracing instantly when starting up gen_server as it is already done which seems to work in most cases. But as a backoff for any application that needs more time the same `TracingManager` gen_server waits for telemetry event `[:phoenix, :live_view, :mount, :start]` of debugged application and restarts tracing after first occurrence and then detaches from telemetry.

### DIFF
--- a/lib/live_debugger/services/callback_tracer/actions/tracing.ex
+++ b/lib/live_debugger/services/callback_tracer/actions/tracing.ex
@@ -15,23 +15,17 @@ defmodule LiveDebugger.Services.CallbackTracer.Actions.Tracing do
 
   @spec setup_tracing!(TracingManager.state()) :: pid() | nil
   def setup_tracing!(state) do
-    pid =
-      case Dbg.tracer({&Tracer.handle_trace/2, 0}) do
-        {:ok, pid} ->
-          Process.monitor(pid)
-          pid
+    case Dbg.tracer({&Tracer.handle_trace/2, 0}) do
+      {:ok, pid} ->
+        Process.monitor(pid)
 
-        {:error, error} ->
-          raise "Couldn't start tracer: #{inspect(error)}"
-      end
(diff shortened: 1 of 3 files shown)


lib/live_debugger/services/callback_tracer/actions/tracing.ex
```diff
@@ -12,25 +13,25 @@ defmodule LiveDebugger.Services.CallbackTracer.Actions.Tracing do
...
   alias LiveDebugger.Services.CallbackTracer.Queries.Paths, as: PathQueries
 
-  @spec setup_tracing!() :: :ok
-  def setup_tracing!() do
+  @spec setup_tracing!(TracingManager.state()) :: pid() | nil
+  def setup_tracing!(state) do
     case Dbg.tracer({&Tracer.handle_trace/2, 0}) do
       {:ok, pid} ->
-        Process.link(pid)
+        Process.monitor(pid)
```
hhubert6: Switch to `monitor` in order to be able to restart `dbg` server without restarting whole `gen_server`.

lib/live_debugger/services/callback_tracer/actions/tracing.ex
```diff
@@ -12,25 +13,25 @@ defmodule LiveDebugger.Services.CallbackTracer.Actions.Tracing do
...
-    Dbg.process([:c, :timestamp])
-    apply_trace_patterns()
-
-    :ok
   end
 
   @spec refresh_tracing() :: :ok
   def refresh_tracing() do
-    apply_trace_patterns()
+    Dbg.stop()
```
hhubert6: Stopping and starting `dbg` server again covers more cases in which it might not catch traces than just reapplying trace patterns. For example in Livebook case.

srzeszut (approved):
👌

PR #884, opened 2025-12-22
Enhancement: Add better handling of callbacks exceptions

srzeszut:


lib/live_debugger/services/callback_tracer/gen_servers/trace_handler.ex
```diff
@@ -179,4 +210,42 @@ defmodule LiveDebugger.Services.CallbackTracer.GenServers.TraceHandler do
...
+    {
+      "** (stop) " <> inspect(reason),
+      "(Stacktrace not available)"
+    }
+  end
+
+  defp shorten_message(message) do
+    message
+    |> String.split(~r/\.(\s|$)/, parts: 2)
+    |> List.first()
```
kraleppa: Maybe `trace_utils.ex`?

kraleppa:
Also, I noticed that this gap is much bigger on your branch than on main. Can it be fixed so we use full width of this element?

main: 
Screenshot 2026-01-19 at 11 45 51

this branch:
Screenshot 2026-01-19 at 11 46 21

kraleppa (commented):
I think we're ready to merge after resolving the above issue

hhubert6 (commented):
Also fullscreen of trace with exception should expand the same way as regular one, which is take maximal space.


Image

Image

PR #892, opened 2026-01-07
Feature: Displaying temporary assigns

hhubert6:


lib/live_debugger/app/debugger/node_state/web/components.ex
```diff
@@ -87,52 +87,115 @@ defmodule LiveDebugger.App.Debugger.NodeState.Web.Components do
...
+      assigns
+      |> assign(entries: TermParser.term_to_display_tree(assigns.temporary_assigns).children)
+
+    ~H"""
+    <div id={@id}>
+      <.section_title
+        name={
+          if(Enum.empty?(@temporary_assigns), do: "No temporary assigns", else: "Temporary assigns")
+        }
+        icon="history"
```
kraleppa: I'd change this icon - history is already taken by assigns history. I think on figma we used `clock` icon

PR #898, opened 2026-01-09
Tests: fix flaky tests associated with `select_live_view`

kraleppa:
This should fix or decrease number of flaky tests associated with:
- Finding unexpected number for selector `button[phx-click=\"select-live-view\"]` (usually 1 is expected and 0, 2, 3 could be found)
- Finding unexpected number for selector `#traces-list-stream details` (usually 6 is expected and only 2 are found)


hhubert6 (approved):
Nice 🐬

PR #914, opened 2026-01-25
Set endpoint url config for correct Phoenix startup log

ijcd:
## Problem

When configuring LiveDebugger with a custom IP or `external_url`, the Phoenix startup log still shows:

```
[info] Access LiveDebugger.App.Web.Endpoint at http://localhost:4007
```

## Cause

`put_endpoint_config/1` sets `http: [ip: ..., port: ...]` but not `url: [host: ...]`. Phoenix uses the `url` config for the "Access at" log message, defaulting to `localhost`.

## Fix

Derive `host` from `external_url` if configured, otherwise use the IP string. Add `url: [host: host, port: port]` to endpoint config.

## Result

- Without `external_url`: `Access at http://127.0.0.11:4007`
- With `external_url: "http://my-host.local:4007"`: `Access at http://my-host.local:4007`

## Testing

Tested with Treehouse (dynamic per-branch IP allocation) - both IP fallback and external_url work correctly.

### DIFF
--- a/lib/live_debugger.ex
+++ b/lib/live_debugger.ex
@@ -56,12 +56,19 @@ defmodule LiveDebugger do
   end
 
   defp put_endpoint_config(config) do
+    ip = Keyword.get(config, :ip, @default_ip)
+    port = Keyword.get(config, :port, @default_port)
+
+    host =
+      case Keyword.get(config, :external_url) do
+        nil -> ip |> :inet.ntoa() |> List.to_string()
+        url -> URI.parse(url).host
+      end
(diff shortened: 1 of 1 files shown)


hhubert6 (commented):
Hey, thanks for your contribution. One thing though: `port` of the `external_url` also can be different, so if you can adjust to derive `host` and `port` it'd be nice.

kraleppa:
Closing, because of no response, feel free to re-open after addressing @hhubert6 concerns

PR #916, opened 2026-01-26
Enhancement: Add better indicator of selected node

srzeszut:


lib/live_debugger/app/controllers/socket_discovery_controller.ex
```diff
@@ -18,7 +20,9 @@ defmodule LiveDebugger.App.Controllers.SocketDiscoveryController do
       [lv_process] ->
         conn
         |> Phoenix.Controller.redirect(
-          to: RoutesHelper.debugger_node_inspector(lv_process.pid, node_id)
+          to:
+            RoutesHelper.debugger_node_inspector(lv_process.pid, node_id)
+            |> URL.upsert_query_param("from", from)
```
kraleppa: `from` can be `nil` when opening LiveDebugger via debug button and this function does not accept nil

lib/live_debugger/app/web/components.ex
```diff
@@ -628,30 +629,50 @@ defmodule LiveDebugger.App.Web.Components do
...
   slot(:inner_block)
 
   def sidebar_slide_over(assigns) do
     ~H"""
+    <span
+      :if={@trigger_sidebar}
+      class="
+              hidden
+              [--open-sidebar:1]
+              md_ct:[--open-sidebar:0]"
```
kraleppa:
```suggestion
      class="hidden [--open-sidebar:1] md_ct:[--open-sidebar:0]"
```

PR #919, opened 2026-02-04
Tests: Setup Playwright

hhubert6:


dev/layout.ex
```diff
@@ -75,7 +76,10 @@ defmodule LiveDebuggerDev.Layout do
     ~H"""
     <main class="p-5">
       <.navbar />
-      <p>Current PID: <span id="current-pid"><%= pid_to_string(self()) %></span></p>
+      <p>
+        Current PID:
+        <span :if={connected?(@socket)} id="current-pid"><%= pid_to_string(self()) %></span>
+      </p>
```
hhubert6: When displaying `pid` only when connected there is only one `pid `rendered on the page which makes testing it in e2e tests easier and more reliable.

e2e/playwright.config.ts
```diff
@@ -0,0 +1,94 @@
...
+      use: { ...devices['Desktop Chrome'] },
+      testMatch: '**/*.serial.spec.ts',
+      dependencies: ['chromium', 'firefox'],
+    },
+    {
+      name: 'serial-tests firefox',
+      use: { ...devices['Desktop Firefox'] },
+      testMatch: '**/*.serial.spec.ts',
+      dependencies: ['chromium', 'firefox', 'serial-tests chromium'],
+    },
```
hhubert6: Testing global settings can be isolated from the rest of the tests by adding dependencies between projects. This way all other tests can be run in parallel which speeds up the tests significantly.

e2e/playwright.config.ts
srzeszut: Are those commented sections necessary in this file?

PR #923, opened 2026-02-11
Feature: Add flash message when LiveView crashes

srzeszut:


hhubert6 (commented):
I think it'd be much nicer to redirect to crashed component on the same page instead of opening a new one. Or maybe adding two buttons to give the user a choice. What do you think?

lib/live_debugger/app/debugger/callback_tracing/web/hooks/filter_new_traces.ex
```diff
@@ -61,7 +65,17 @@ defmodule LiveDebugger.App.Debugger.CallbackTracing.Web.Hooks.FilterNewTraces do
...
+          case trace_event do
+            %TraceExceptionUpdated{} = trace_exception ->
+              push_exception_flash(socket, trace_exception)
+
+            _ ->
+              socket
+          end
+
+        {:halt, socket}
     end
```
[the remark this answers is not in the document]
kraleppa: Nice idea!

e2e/tests/exception_trace.spec.ts
```diff
@@ -155,3 +163,29 @@ test('debugger captures runtime errors and exceptions in global callbacks 3', as
...
+  await findErrorButton(devApp, 'crash_argument').click();
+
+  const flash = dbgApp.locator('#flash-error');
+  await expect(flash).toBeVisible();
+  await expect(flash).toContainText('LiveComponent crashed.');
+  await expect(flash).toContainText('LiveDebuggerDev.LiveComponents.Crash');
+  await expect(flash).toContainText('Open in Node Inspector');
+});
+
+test("flash message doesn't appear when node crashes", async ({
```
kraleppa: Maybe you could extend this case a bit to cover cases with filters and search phrase. These are e2e tests and should be broader - "flash message doesn't appear when node crashes" seemed like a name of unit frontend test.

I'd implement here two tests: one covering a case when you have an error in a different node, and one when you have the error in currently debugged node, but with all these extra stuff

lib/live_debugger/app/web/hooks/flash.ex
```diff
@@ -19,23 +34,26 @@ defmodule LiveDebugger.App.Web.Hooks.Flash do
...
           socket :: Phoenix.LiveView.Socket.t(),
           key :: String.t() | atom(),
-          message :: String.t()
+          message :: String.t() | ExceptionFlashData.t()
         ) ::
           Phoenix.LiveView.Socket.t()
   def push_flash(socket, key, message)
-      when is_binary(message) and key in [:error, :info] do
+      when key in [:error, :info] and
+             (is_binary(message) or is_struct(message, ExceptionFlashData)) do
```
kraleppa: I wouldnt do such strong guarding here tbh, checking keys via guard is enough

PR #966, opened 2026-04-03
Add tour infrastructure

srzeszut:


lib/live_debugger/api/settings_storage.ex
hhubert6, line 108: I think the only change needed in this file is not using env config so the user won't overwrite defaults. This  and `:settings_enabled` flag in settings page should be enough.
```suggestion
|> Enum.map(fn setting ->
{setting, fetch_setting(setting)}
end)
```

lib/live_debugger/app/settings/web/settings_live.ex
```diff
@@ -194,5 +211,20 @@ defmodule LiveDebugger.App.Settings.Web.SettingsLive do
     |> noreply()
   end
 
+  def handle_info({"tour:settings-enabled", _payload}, socket) do
```
hhubert6, line 219: As far as I can see all events from tour api have message `"tour:action"` so this callback won't work. You either need to much the payload `%{"action" => "settings-enabled"}` or change the way messages are sent.

lib/live_debugger/tour_elements.ex
hhubert6, line 1: This file should be rather in `lib/live_debugger` directory

lib/live_debugger/app/discovery/web/discovery_live.ex
hhubert6, line 1: `DiscoveryLive` is missing `phx-hook="Tour"` and `handle_event` for `step-completed`. Also I think it's a good idea to extract those two tour callbacks to an external hook.

lib/live_debugger/app/settings/web/settings_live.ex
hhubert6, line 1: IMO having usual tour callbacks for spotlights might be useful.

assets/app/hooks/tour.js
```diff
@@ -0,0 +1,181 @@
...
+    const handler = () => {
+      clearAll();
+      this._cleanup = null;
+    };
+
+    setTimeout(() => {
+      document.addEventListener('click', handler, { once: true });
+      this._cleanup = () => document.removeEventListener('click', handler);
+    }, 0);
+  },
```
hhubert6, line 180: "click anywhere" action should also send "step-completed" event just in case

hhubert6 (commented):
There is an issue with scroll and spotlight. When spotlighted element is out of view because of scroll whole page become frozen. Adding auto scroll to spotlighted element should fix that (and hightlight for consistency).

hhubert6 (approved):
🐬 please update assets before merging

PR #969, opened 2026-04-09
Add info to the remaining node inspector sections

### LINKED ISSUE #805: Consider renaming "callback tracing"
Name callback tracing does not describe very well this feature. To user who is not familiar with tracing this name may be misleading. We should consider changing its name to e.g. "LiveView/LiveComponent lifecycle"

srzeszut:


kraleppa (approved):

hhubert6 (approved):
You can move info icon of temporary assigns to the left as well

PR #972, opened 2026-04-13
Fix: disable tracing when heap size limit is reached

GuzekAlan:


lib/live_debugger/services/callback_tracer/gen_servers/tracing_manager.ex
```diff
@@ -74,9 +77,9 @@ defmodule LiveDebugger.Services.CallbackTracer.GenServers.TracingManager do
 
   # handling dbg tracer stop
   def handle_info({:DOWN, _, _, pid, :done}, %{dbg_pid: pid} = state) do
-    state = TracingActions.setup_tracing_with_monitoring!(state)
+    Bus.broadcast_event!(%DbgKilled{})
 
-    {:noreply, state}
+    {:noreply, %{state | dbg_pid: nil}}
   end
```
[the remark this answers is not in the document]
GuzekAlan: `:dbg` traps exit messages and always returns `:done` https://github.com/erlang/otp/blob/b40d9ef70d87747cdf8c0eeb9b818fafa2fd0108/lib/runtime_tools/src/dbg.erl#L1951-L1964

lib/live_debugger/utils/memory.ex
```diff
@@ -28,6 +28,21 @@ defmodule LiveDebugger.Utils.Memory do
...
+  @spec set_max_heap_size(size :: number()) :: :ok
+  def set_max_heap_size(size) when is_number(size) do
+    Process.flag(:max_heap_size, %{
+      kill: true,
+      error_logger: false,
+      size: trunc(size * @gigabyte / @wordsize)
+    })
+
+    :ok
+  end
```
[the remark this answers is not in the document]
GuzekAlan: It would only test `Process.flag` function which is unnecessary

lib/live_debugger/app/web/components/tracing_crash_popup.ex
```diff
@@ -0,0 +1,92 @@
...
+  defp handle_info(_, socket), do: {:cont, socket}
+
+  defp tracing_started?(tracing_status) when tracing_status in @tracing_statuses do
+    case tracing_status do
+      :started -> true
+      _ -> false
+    end
+  end
+
+  defp initial_status(started?), do: if(started?, do: :started, else: :stopped)
```
kraleppa:
```suggestion
  defp initial_status(true), do: :started
  defp initial_status(false), do: :stopped
```

lib/live_debugger/services/callback_tracer/gen_servers/trace_handler.ex
```diff
@@ -52,6 +56,9 @@ defmodule LiveDebugger.Services.CallbackTracer.GenServers.TraceHandler do
 
   @impl true
   def init(_opts) do
+    Bus.receive_events!()
+    Memory.set_max_heap_size(@max_heap_size)
```
kraleppa, line 58: Trace handler can be killed, but theoretically `tracer` process can still be working right? There is a chance that handler will have bigger memory usage than tracer - what would happen in such case?

lib/live_debugger/services/callback_tracer/gen_servers/trace_handler.ex
```diff
@@ -52,6 +55,9 @@ defmodule LiveDebugger.Services.CallbackTracer.GenServers.TraceHandler do
 
   @impl true
   def init(_opts) do
+    Process.flag(:trap_exit, true)
```
kraleppa: Why do we do this? We trap the exit signal of the linked processes and then we catch them in the handle_info in 207 line which stops the GenServer gracefully, right? What's the point?

GuzekAlan: Initially I wanted to capture exit, since it should be `:killed` but after more reading it seems that `max_heap_size` sends `untrappable` exit so there is no point in trapping it. It restarts normally and from what I've tested works fine without it

docs/config.md
```diff
@@ -24,6 +24,18 @@ In `router.ex` of your Phoenix app, make sure your locally running Phoenix app c
...
+There are 2 memory consuming processes which are running in the background `Tracer` and `TraceHandler`. You can set maximum size of heap (in Gigabytes) which these processes can take:
+
+```elixir
+# config/dev.exs
+
+config :live_debugger, 
+  tracer_max_heap_size: 5,
+  trace_handler_max_heap_size: 5
+```
+
```
hhubert6, line 38: It'd be best to explicitly say what are default values for this config

hhubert6 (commented):
I think the popup should explain a little bit more about what could be possible causes of tracer crashing

lib/live_debugger/app/web/hooks/tracer_status.ex
```diff
@@ -0,0 +1,78 @@
...
+
+  defp handle_info(%DbgStarted{}, socket) do
+    {:halt, assign(socket, :tracer_started?, AsyncResult.ok(true))}
+  end
+
+  defp handle_info({:tracer_status, :refetch}, socket) do
+    socket
+    |> assign(:tracer_started?, AsyncResult.loading())
+    |> start_async(:tracer_started?, fn -> tracer_status() end)
+  end
```
hhubert6: There is missing `{:halt, socket}` here

lib/live_debugger/app/settings/web/settings_live.ex
```diff
@@ -20,11 +22,18 @@ defmodule LiveDebugger.App.Settings.Web.SettingsLive do
...
 
   @impl true
-  def handle_params(params, _url, socket) do
+  def mount(_params, _session, socket) do
     if connected?(socket) do
       Bus.receive_events!()
     end
 
+    socket
+    |> TracerStatusHook.init()
```
hhubert6, line 31: This hook also does `Bus.receive_events!()` so it duplicates events this LiveView receives. Same goes for `DiscoveryLive` and `DebuggerLive`

lib/live_debugger/app/web/hooks/tracer_status.ex
```diff
@@ -0,0 +1,71 @@
...
+
+  alias LiveDebugger.Services.CallbackTracer.GenServers.TracingManager
+  alias Phoenix.LiveView.AsyncResult
+
+  alias LiveDebugger.Services.CallbackTracer.Events.DbgKilled
+  alias LiveDebugger.Services.CallbackTracer.Events.DbgStarted
+
+  require Logger
+
+  @tracer_state_request_timeout 3000
```
GuzekAlan, line 20: 10s looked like too much of a timeout. What do you think ?

lib/live_debugger/app/web/live_components/tracer_status.ex
```diff
@@ -0,0 +1,173 @@
...
+  defp maybe_reset_on_started(socket, _), do: socket
+
+  attr(:restarting?, :boolean, required: true)
+  attr(:myself, :any, required: true)
+
+  defp tracer_crash_info(assigns) do
+    ~H"""
+    <div class="bg-error-bg border-b border-error-border text-error-text flex items-center justify-center py-2 px-4 text-xs">
+      <.icon name="icon-exclamation-circle" class="w-3 h-3 inline-block mr-1 align-middle" />
+      <b>Tracer crashed</b>
```
kraleppa: You'e missing whitepsace before `-`
Image

lib/live_debugger/app/web/live_components/tracer_status.ex
```diff
@@ -0,0 +1,173 @@
...
+      - restart to enable core functionalities
+      <button
+        phx-click="restart_tracing"
+        phx-target={@myself}
+        disabled={@restarting?}
+        class="inline-flex items-center gap-1.5 ml-2 rounded text-xs font-semibold py-1.5 px-2 bg-error-bg border border-error-border text-button-red-content hover:bg-error-border hover:text-button-red-content-hover disabled:opacity-50 disabled:pointer-events-none"
+      >
+        <.spinner :if={@restarting?} size="xs" /> Restart Tracing
+      </button>
```
kraleppa, line 99: This button looks a bit bad, on the light mode (on dark mode it's fine)
Image

kraleppa (commented):
If I dismiss the modal, and change page the modal appears again. Imo after you dismiss it once, it should not appear again
https://github.com/user-attachments/assets/12de7d20-1796-4daa-9eda-37a2e810c406

lib/live_debugger/app/web/live_components/tracer_status.ex
```diff
@@ -0,0 +1,181 @@
...
+        </div>
+
+        <div class="px-3">
+          Core functionalities like <b>Assigns</b>
+          and <b>Callback Tracing</b>
+          won't have newest information and may be misleading.
+        </div>
+        <div class="px-3">
+          This may be caused by <b>debugged application</b>
+          having an excessive amount of traces with huge sizes - LiveDebugger has a max heap size set for tracing. See
```
hhubert6:
```suggestion
          having an excessive number of traces with huge sizes - LiveDebugger has a max heap size set for tracing. See
```

kraleppa (approved):

PR #976, opened 2026-04-20
Fix: highlighting LiveViews by their children bounding boxes

### LINKED ISSUE #973: Highlighting LiveViews not working
In particular projects highlighting LiveViews via Components Tree, Associated LiveViews or inspecting seems not to work, highlighting LiveComponents works fine.
Example repo where it happens: https://github.com/software-mansion-labs/live-debugger-tour

GuzekAlan:


assets/client/services/highlight.js
```diff
@@ -88,82 +94,84 @@ function handleHighlight({ detail }, shadowRoot) {
     `[${detail.attr}="${detail.val}"]`
   );
 
-  if (isElementVisible(activeElement)) {
-    highlightElement = createHighlightElement(
-      activeElement,
-      detail,
-      highlightElementID
-    );
+  if (!activeElement) return;
```
kraleppa, line 97: You've removed `isElementVisible` - why? What if we wanted to highlight hidden element?

GuzekAlan: Since this bugfix was to show bounding boxes for LiveViews and LiveComponents I thought that we might want to always show them and highlight. I can add it back and not highlight if it is not visible with expect of the boilerplate scenario from `phx.new`. Question is if there won't be an example where users will define LiveComponent or LiveView wrappers as not visible and we won't detect it. What do you think? 

I've tried at first to leave the `isVisible` check but the bug was still there then

kraleppa: ok, I guess we can leave it as it is

kraleppa (approved):

srzeszut (approved):

PR #988, opened 2026-05-10
Fix MapSet assign summary updates

### LINKED ISSUE #986: MapSet struct doesn't update
### Bug Description

In `Assigns` tab if there is a `MapSet` which changes values the collapsed view shows only initial ones.

Image

Image

After LiveDebugger page reload MapSet shows correct values but STILL doesn't update

Image

### Actual Behavior

MapSet structs in `Assigns` don't update on collapsed view

### Expected Behavior

The values should be updated for both collapsed and open views

### Steps to Reproduce

Example behaviour in https://github.com/software-mansion-labs/live-debugger-tour .
1. run tour and go to any tutorial page when debugger is open
2. complete some steps (click `Spotlight`)
3. check `completed_steps` assign in LiveDebugger

Example LiveView module with this behaviour:

```elixir
defmodule NewPhoenixAppWeb.HelloLive do
  use NewPhoenixAppWeb, :live_view

  def mount(_params, _session, socket) do
    {:ok, assign(socket, message: MapSet.new([]))}
  end

  def render(assigns) do
    ~H"""
    <div class="px-4 py-10 sm:px-6 sm:py-28 lg:px-8 xl:px-28 xl:py-32">
      <div class="mx-auto max-w-xl">
        <h1 class="text-4xl font-bold tracking-tight">Live View</h1>
        <p class="mt-4 text-xl text-base-content/70">{Enum.map(@message, fn e -> inspect(e) end)}</p>
        <button phx-click="add">Add</button>
      </div>
    </div>
    """
  end

  def handle_event("add", _, socket) do
    message  = socket.assigns.message
    size = MapSet.size(message) |> Integer.to_string()
    message = MapSet.put(message, size)

    {:noreply, assign(socket, :message, message)}
  end
end

```

### Elixir version with Erlang/OTP

Elixir 1.19.1 OTP 28

### LiveDebugger version

1.0.0

### Phoenix version

1.8.5

### Phoenix LiveView version

1.1.30

### Operating system

MacOS 26.3.1

### Browser(s)

Brave

dl-alexandre:
## Summary

- Carry the full updated struct through struct diffs so display nodes can refresh value-dependent collapsed content.
- Refresh struct display content while preserving existing child diffs and expanded state.
- Add a regression test for a `MapSet` assign whose collapsed summary changes after a diff update.

Fixes #986

## Validation

- `mix format --check-formatted lib/live_debugger/app/utils/term_differ.ex lib/live_debugger/app/utils/term_parser.ex test/app/utils/term_differ_test.exs test/app/utils/term_parser_test.exs`
- `mix test test/app/utils/term_differ_test.exs test/app/utils/term_parser_test.exs test/app/debugger/node_state/queries_test.exs test/app/debugger/node_state/web/components_test.exs`
- `mix test`

### DIFF
--- a/lib/live_debugger/app/utils/term_differ.ex
+++ b/lib/live_debugger/app/utils/term_differ.ex
@@ -100,7 +100,6 @@ defmodule LiveDebugger.App.Utils.TermDiffer do
     %Diff{
       type: :struct,
       ins: %{@primitive_key => struct2},
-      del: %{@primitive_key => struct1},
       diff: diff
     }
   end
--- a/lib/live_debugger/app/utils/term_differ.ex
+++ b/lib/live_debugger/app/utils/term_differ.ex
@@ -97,7 +97,12 @@ defmodule LiveDebugger.App.Utils.TermDiffer do
     map2 = Map.from_struct(struct2)
     {_, _, diff} = map_diffs(map1, map2)
 
-    %Diff{type: :struct, diff: diff}
+    %Diff{
+      type: :struct,
+      ins: %{@primitive_key => struct2},
+      del: %{@primitive_key => struct1},
+      diff: diff
+    }
   end
 
   def diff(struct1, struct2) when is_struct(struct1) and is_struct(struct2) do
--- a/lib/live_debugger/app/utils/term_parser.ex
+++ b/lib/live_debugger/app/utils/term_parser.ex
@@ -242,37 +242,30 @@ defmodule LiveDebugger.App.Utils.TermParser do
          %Diff{ins: %{@primitive_key => struct}},
          opts
        ) do
+     {content, expanded_before, expanded_after} = struct_display_elements(struct)
+
+    new_node =
+      TermNode.new(:struct, content,
+        expanded_before: expanded_before,
+        expanded_after: expanded_after
+      )
+
     new_node =
       case Keyword.get(opts, :key) do
         nil ->
-          {content, expanded_before, expanded_after} = struct_display_elements(struct)
-
-          TermNode.new(:struct, content,
-            expanded_before: expanded_before,
-            expanded_after: expanded_after
-          )
+          new_node
 
         key ->
           {key_span, sep_span} = key_prefix_and_separator(key)
-          {content, expanded_before, expanded_after} = struct_display_elements(struct)
 
-          summary =
-            TermNode.new(:struct, content,
-              expanded_before: expanded_before,
-              expanded_after: expanded_after
-            )
-
-          prefixed = TermNode.add_prefix(summary, [key_span, sep_span])
-          %TermNode{prefixed | key: key}
+          TermNode.add_prefix(new_node, [key_span, sep_span])
       end
 
     %TermNode{
       term_node
       | content: new_node.content,
         expanded_before: new_node.expanded_before,
-        expanded_after: new_node.expanded_after,
-        kind: new_node.kind,
-        key: new_node.key
+        expanded_after: new_node.expanded_after
     }
   end
 
--- a/lib/live_debugger/app/utils/term_parser.ex
+++ b/lib/live_debugger/app/utils/term_parser.ex
@@ -8,6 +8,8 @@ defmodule LiveDebugger.App.Utils.TermParser do
   alias LiveDebugger.App.Utils.TermNode.DisplayElement
   alias LiveDebugger.App.Utils.TermNode
 
+  @primitive_key TermDiffer.primitive_key()
+
   @doc """
   Convert term into infinite string which can be copied to IEx console.
   """
@@ -213,9 +215,7 @@ defmodule LiveDebugger.App.Utils.TermParser do
   end
 
   defp refresh_struct_content(%TermNode{} = term_node, %Diff{ins: ins}, opts) do
-    primitive_key = TermDiffer.primitive_key()
-
-    case Map.fetch(ins, primitive_key) do
+    case Map.fetch(ins, @primitive_key) do
       {:ok, struct} ->
         new_node =
           case Keyword.get(opts, :key) do
--- a/lib/live_debugger/app/utils/term_parser.ex
+++ b/lib/live_debugger/app/utils/term_parser.ex
@@ -214,33 +214,60 @@ defmodule LiveDebugger.App.Utils.TermParser do
     end
   end
 
-  defp refresh_struct_content(%TermNode{} = term_node, %Diff{ins: ins}, opts) do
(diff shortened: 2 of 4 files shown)


lib/live_debugger/app/utils/term_differ.ex
```diff
@@ -97,7 +97,12 @@ defmodule LiveDebugger.App.Utils.TermDiffer do
     map2 = Map.from_struct(struct2)
     {_, _, diff} = map_diffs(map1, map2)
 
-    %Diff{type: :struct, diff: diff}
+    %Diff{
+      type: :struct,
+      ins: %{@primitive_key => struct2},
+      del: %{@primitive_key => struct1},
```
GuzekAlan: `del` seems to be unnecessary and it will use more memory. Consider adding only `ins` or even storing the `inspect` of the struct

lib/live_debugger/app/utils/term_parser.ex
```diff
@@ -210,6 +212,36 @@ defmodule LiveDebugger.App.Utils.TermParser do
     end
   end
 
+  defp refresh_struct_content(%TermNode{} = term_node, %Diff{ins: ins}, opts) do
+    primitive_key = TermDiffer.primitive_key()
```
GuzekAlan: Nitpick: use it as `module attribute`

dl-alexandre:
Thanks for the review @GuzekAlan — both points addressed in 4ca82c9:

- **`term_differ.ex`** — dropped `del` from the `:struct` diff. `TermParser.refresh_struct_content/3` only reads `ins`, so carrying the old struct was unused memory. Kept the `:struct` shape minimal (`ins` + `diff`).
- **`term_parser.ex`** — `primitive_key` is now a `@primitive_key` module attribute instead of a per-call `TermDiffer.primitive_key/0`.
- Updated the `term_differ_test.exs` struct-diff expectation to match.

Compiles clean with `--warnings-as-errors`.

lib/live_debugger/app/utils/term_parser.ex
```diff
@@ -210,6 +214,34 @@ defmodule LiveDebugger.App.Utils.TermParser do
     end
   end
 
+  defp refresh_struct_content(%TermNode{} = term_node, %Diff{ins: ins}, opts) do
+    case Map.fetch(ins, @primitive_key) do
```
GuzekAlan: This might be done in a pattern match to avoid nesting

lib/live_debugger/app/utils/term_parser.ex
```diff
@@ -210,6 +214,34 @@ defmodule LiveDebugger.App.Utils.TermParser do
...
+      {:ok, struct} ->
+        new_node =
+          case Keyword.get(opts, :key) do
+            nil ->
+              to_node(struct)
+
+            key ->
+              {key, struct}
+              |> to_key_value_node()
+              |> elem(1)
```
GuzekAlan: Is recalculating all children necessary if we want to only update the content?

dl-alexandre:
Addressed the latest review comments:

- Refactored `refresh_struct_content/3` to use direct pattern matching on the `%Diff{ins: %{@primitive_key => struct}}` clause (plus a catch-all), eliminating the nested `case Map.fetch/2`.
- Extracted `struct_summary_node/1` which builds a TermNode containing only the collapsed `content`/`expanded_before`/`expanded_after`/`kind` (with empty children) so that summary refreshes for structs like `MapSet` no longer unnecessarily recalculate child nodes via `to_key_value_children/1`.
- Extracted `key_prefix_and_separator/1` to share the key-label logic and avoid duplication.

- `mix format`, targeted tests (`term_*_test.exs`, node_state tests), and `--warnings-as-errors` all pass.

Ready for re-review / merge. This keeps the child diff/expansion state intact while fixing the collapsed summary for changing `MapSet` (and other custom-Inspect structs) assigns.

lib/live_debugger/app/utils/term_parser.ex
```diff
@@ -210,6 +214,61 @@ defmodule LiveDebugger.App.Utils.TermParser do
...
+        [DisplayElement.black(inspect(struct))]
+      end
+
+    expanded_before = [
+      DisplayElement.black("%"),
+      DisplayElement.blue(inspect(module)),
+      DisplayElement.black("{")
+    ]
+
+    expanded_after = [DisplayElement.black("}")]
```
GuzekAlan, line 235: This logic is same in `to_node`. Extract it as a separate function and return `expand_after`, `expand_before` and `content` (not `TermNode`). Use it inside `to_node` and `refresh_struct_content`

test/app/utils/term_parser_test.exs
```diff
@@ -748,6 +748,23 @@ defmodule LiveDebugger.App.Utils.TermParserTest do
...
+
+      assert {:ok, updated_node} = TermParser.update_by_diff(term_node, diff)
+      {_, message_node} = List.keyfind(updated_node.children, :message, 0)
+
+      assert Enum.map(message_node.content, & &1.text) == [
+               "message:",
+               " ",
+               "MapSet.new([\"0\"])"
+             ]
+    end
```
GuzekAlan: please add more tests to properly check structs inside maps, lists and other structs. Use more nesting in tests to check if it updates properly without touching other elements

lib/live_debugger/app/utils/term_parser.ex
```diff
@@ -210,6 +214,70 @@ defmodule LiveDebugger.App.Utils.TermParser do
...
+      end
+
+    %TermNode{
+      term_node
+      | content: new_node.content,
+        expanded_before: new_node.expanded_before,
+        expanded_after: new_node.expanded_after,
+        kind: new_node.kind,
+        key: new_node.key
+    }
```
GuzekAlan, line 269:
```suggestion
     {content, expanded_before, expanded_after} = struct_display_elements(struct)

    new_node =
      TermNode.new(:struct, content,
        expanded_before: expanded_before,
        expanded_after: expanded_after
      )

    new_node =
      case Keyword.get(opts, :key) do
        nil ->
          new_node

        key ->
          {key_span, sep_span} = key_prefix_and_separator(key)

          TermNode.add_prefix(new_node, [key_span, sep_span])
      end

    %TermNode{
      term_node
      | content: new_node.content,
        expanded_before: new_node.expanded_before,
        expanded_after: new_node.expanded_after
    }
```

GuzekAlan: `kind` won't change and there was duplicated code

GuzekAlan (approved):

PR #989, opened 2026-05-12
Enhancement: Improve tracing performance

### LINKED ISSUE #975: Improve tracing performance
We received some time this issue https://github.com/software-mansion/live-debugger/issues/956 and we kind of solved it by adding memory limit for risky processes here https://github.com/software-mansion/live-debugger/pull/972

The goal here is to research possible improvements for tracing that would decrease memory usage on heavy LiveViews

@GuzekAlan already did some research - please write in the comments your findings

hhubert6:
I've tried several approaches to improve performance, but none of them yielded a significant enough impact to comfortably handle the `MemoryExplosion` examples included in the dev app.

Some of the things I tried that give noticeable difference:
- Using file port in `:dbg.tracer `with `:dbg.trace_client`
	- `:follow_file` creates challenges of manual file size management (I haven't gone deeper on that)
	- `:file` with `FileWrapSpec` manages file size well but does not support streaming traces live so it becomes really problematic with iterative traces fetching
- Using ip port in `:dbg.tracer `with `:dbg.trace_client`
- Replacing `:dbg.tracer` with plain `:erlang.trace`
- Using custom erl_tracer NIF for reducing trace size by dropping some field like socket from assigns in `render/1` callback or assigns from socket in other callback (e.g. `handle_event/3`)
- Dynamic spawning of TraceHandler GenServer for load balancing of incoming traces

All listed options give somewhat noticeable difference in performance and/or memory usage but each potentially creates known challenges or ones that may occur later. That's why I went with simple replacement of `:dbg.tracer` with `:erlang.trace` as it's least prone to unknown issues, the easiest to manage and gives visible increase in performance.

Based on what I've researched, attempting to improve performance beyond that would require fundamental and extensive redesign of the entire tracing logic, likely involving different design assumptions than those we initially made.


lib/live_debugger/services/callback_tracer/actions/tracing.ex
```diff
@@ -31,7 +31,10 @@ defmodule LiveDebugger.Services.CallbackTracer.Actions.Tracing do
       {:ok, pid} ->
         Process.monitor(pid)
 
-        Dbg.process([:c, :timestamp, :procs])
+        case Dbg.process([:c, :timestamp]) do
+          {:ok, _} -> :ok
+          {:error, error} -> raise "Couldn't enable system-wide call tracing: #{inspect(error)}"
```
kraleppa: This error will probably completely crash LiveDebugger on start - can't we handle this the same way as we handle cases when `tracer` process dies? We let LiveDebugger start but we display this popup with info that tracing is down, and please refresh it.

kraleppa (approved):
Very nice!

kraleppa (approved):
nice 🐬

GuzekAlan (approved):
LGTM 🦈

PR #997, opened 2026-07-06
Fix recurring DETS "not properly closed" log from settings storage

### LINKED ISSUE #996: SettingsStorage  never closes its DETS table → "not properly closed, repairing" logged on almost every start
### Bug Description

`LiveDebugger.API.SettingsStorage.Impl.init/0` (lib/live_debugger/api/settings_storage.ex ) opens a DETS table at `_build/<env>/lib/live_debugger/live_debugger_saved_settings` and never closes it. It's opened from  `LiveDebugger.start/2  →  get_children/0` ; nothing calls `:dets.close/1`, and no supervised process owns the table with a `terminate/2`.

On VM exit the file is left in the "open" state, so the next `:dets.open_file/2`  auto-repairs and logs:

```
[info] dets: file ~c".../live_debugger_saved_settings" not properly closed, repairing ...
```

Since it's left dirty on every exit, the line reappears on nearly every start. Data is not lost (DETS repairs it); the log is just noisy and recurring.

Most visible with short-lived Mix tasks: bare  mix  runs the  run  task, which starts the app then hard-halts the VM — no  Application.stop/1  or  terminate/2  runs, so DETS is never closed. Also occurs on Ctrl+C shutdown of a dev server.

## Root cause

- `:dets.close/1`  is never called. It can't be reliably added via a shutdown hook:  `mix run /bare  mix  hard-halts the VM (no  terminate / stop  callbacks run), and  
- `:dets.close/1`  may only be called by the process that opened the table (otherwise  `{:error, :not_owner}` ).

## Possible fix

Back settings with ETS + a plain file (matching  `TracesStorage` / `StatesStorage`+ ), which has no open/dirty flag — or, minimally, open/close the DETS table per operation. Related: #902 (settings desync in DETS).

### Actual Behavior

Logs an `[info]` on every `mix` call, which should be silent.

### Expected Behavior

Properly close DETS, which sometimes cannot happen

### Steps to Reproduce

1. Add  {:live_debugger, "~> 1.0", only: :dev}  to a Phoenix app.
2. Run  mix run -e ":ok"  twice (bare  mix  also works).
3. Second run logs  dets: file ... not properly closed, repairing ... .

### Elixir version with Erlang/OTP

Elixir 1.20.2, Erlang/OTP 29 (erts 17.0.2)

### LiveDebugger version

1.0.1 (same code on  main  @ c5b4d63)

### Phoenix version

1.8.8 (phoenix) f0c84303

### Phoenix LiveView version

1.2.5 (phoenix_live_view) 864e71ee

### Operating system

Arch

### Browser(s)

n.a.

chgeuer:
Fixes #996.

Simply speaking, whenever I run `mix` multiple times on a freshly compiled project, I expect 'silence' each time, but I get this:

```
laptop my_project master  ? ❯ mix

laptop my_project master  ? ❯ mix
[info] dets: file ~c"/home/user/my_project/_build/dev/lib/live_debugger/live_debugger_saved_settings" not properly closed, repairing ...

laptop my_project master  ? ❯ 
```

## Problem
`SettingsStorage` opens a DETS table in `init/0` (via `LiveDebugger.start/2` → `get_children/0`) and never closes it. On VM exit the file is left in the "open" state, so the next `:dets.open_file/2` auto-repairs and logs `dets: file ... not properly closed, repairing ...` on nearly every start. Data is intact, but the log is noisy and recurring. Full details in #996.

## Why not just close on shutdown
`:dets.close/1` can't be added reliably via a shutdown hook:
- `mix run` / bare `mix` hard-halts the VM — no `terminate/2` or `Application.stop/1` runs.
- `:dets.close/1` may only be called by the process that opened the table (`{:error, :not_owner}` otherwise).

## Change
Back the settings with an ETS table (fast reads, same pattern as `TracesStorage` and `StatesStorage`) plus an atomically-written plain term file for persistence. A plain file has no open/dirty flag, so an abrupt halt can never trigger a repair.

- Public API, `@behaviour` callbacks, and `config > saved > default` precedence are unchanged.
- Only values changed at runtime via `save/2` are persisted, so a value set in config keeps winning on every boot.
- An old DETS file at the same path is ignored (decode failure → defaults) and superseded on the first `save/2`.

## Verification
- Bare `mix` 3×: no repair log.
- `save/2` writes the term file; a fresh VM boot reloads the saved value (persistence across restarts intact).
- Changed file compiles with no warnings.

Alternative considered: keep DETS and open/close per operation (smaller diff). Went with ETS + file to match the sibling storages and avoid per-read file I/O — happy to switch to the minimal DETS variant if you prefer.

### DIFF
--- a/lib/live_debugger/api/settings_storage.ex
+++ b/lib/live_debugger/api/settings_storage.ex
@@ -189,15 +189,14 @@ defmodule LiveDebugger.API.SettingsStorage do
     # platforms (notably on Windows it returns `{:error, :eexist}`), which would
     # make every `save/2` after the first one fail and silently stop persisting.
     # Remove the stale file and retry so persistence keeps working across saves.
+    #
+    # Note: the first `with` clause matches the *error* case on purpose, so any
+    # other result (including the happy-path `:ok`) is returned as-is. Do not add
+    # an `else` clause here or those results would be routed to it.
     defp rename_over(tmp, path) do
-      case File.rename(tmp, path) do
-        {:error, :eexist} ->
-          with :ok <- File.rm(path) do
-            File.rename(tmp, path)
-          end
-
-        other ->
-          other
+      with {:error, :eexist} <- File.rename(tmp, path),
+           :ok <- File.rm(path) do
+        File.rename(tmp, path)
       end
     end
 
--- a/lib/live_debugger/api/settings_storage.ex
+++ b/lib/live_debugger/api/settings_storage.ex
@@ -12,9 +12,13 @@ defmodule LiveDebugger.API.SettingsStorage do
   API for managing settings storage. In order to properly use invoke `init/0` at the start of application.
   Settings are kept in an ETS table for fast access and persisted to a file
   (inside `_build/*/live_debugger/`) so they survive application restarts.
-  Settings are retrieved in this order:
-  1. locally saved file (inside `_build/*/live_debugger/` directory)
-  2. default values
+  On `init/0` each setting value is resolved with the following precedence:
+  1. value set in the application config (`config :live_debugger, <setting>, ...`)
+  2. value persisted to the local file (inside `_build/*/live_debugger/` directory)
+  3. default value
+
+  Only values changed at runtime via `save/2` are persisted, so a value set in
+  config keeps taking precedence on every application start.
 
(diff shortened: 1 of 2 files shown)


kraleppa (approved):
Nice, thanks for the PR!

